PR #3, opened 2019-01-20
[MPOM-209]. Corrected Jira URL.

ofarganc:
Thanks for the comments @hboutemy ! I also think this URL is cleaner this way without the whitespaces. I'd prefer to keep the %3D just in case.


hboutemy:
ok, let's remain strict: the current url remains quite easy to understand, now that the %20 are removed

thank you, I knew I had to do something but never had the energy to really do it without some help

PR #20, opened 2020-06-16
[MPOM-251] update minimum version to 3.1.1

elharo:
@hboutemy @rfscholte

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -355,7 +355,7 @@ under the License.
             <configuration>
               <rules>
                 <requireMavenVersion>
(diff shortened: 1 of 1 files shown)


kwin:
Can we please make it overwritable via property (https://issues.apache.org/jira/browse/MPOM-247) and only use 3.1.1 as new default?

elharo:
That's a separate, non-blocking issue, about which I don't have a strong opinion one way or the other. You can bring it up on the dev list and see what others think.

PR #29, opened 2021-01-03
[MPOM-247] make minimum Maven version configurable via property

kwin:
"mavenVersion"

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,7 @@ under the License.
     <sourceReleaseAssemblyDescriptor>source-release</sourceReleaseAssemblyDescriptor>
     <gpg.useagent>true</gpg.useagent>
     <arguments />
+    <minimalMavenBuildVersion>3.1.1</minimalMavenBuildVersion>
     <maven.compiler.source>1.7</maven.compiler.source>
     <maven.compiler.target>1.7</maven.compiler.target>
     <surefire.version>2.22.2</surefire.version>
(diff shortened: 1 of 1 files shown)


slachiewicz (approved):

michael-o (approved):

rfscholte:
There are 2 different meanings for mavenVersion:
- The version required to build the project
- The maven compatible version

We use the property for the latter, i.e. using it for dependency versions.

It is very good possible to require Maven4 to create a Maven 3.1 compatible plugin.

So I'm -1

kwin:
The property name `mavenVersion` was suggested by @slachiewicz, but I am open for other suggestions to make the intention clearer. What about `mavenRequiredVersionForBuild`?

rfscholte:
Just a proposal: `<minimalMavenBuildVersion>${mavenVersion}</minimalMavenBuildVersion>`
Still quite a long name, though

kwin:
@rfscholte Do you have any examples of usages of `mavenVersion`? I would like to add some comment to make the differences between the two properties clearer. At least in ASF Parent there is no usage of `mavenVersion`.

rfscholte:
It is not done at any parent level, but on plugin level. The reason is that prerequisites are not inherited, so you need to define them per plugin.
But that does make my proposal awkward, I see now.

kwin:
I now use `minimalMavenBuildVersion`, but give it a literal default value, as property `mavenVersion` is not defined in this scope.

kwin:
@rfscholte Are you fine with this PR?

rfscholte (approved):

jira-importer:
Resolve #431

PR #30, opened 2021-01-03
MPOM-255 enforce setting property "project.build.outputTimestamp"

kwin:


pom.xml
```diff
@@ -330,6 +332,13 @@ under the License.
                 <requireMavenVersion>
                   <version>3.1.1</version>
                 </requireMavenVersion>
+                <requireProperty>
+                  <property>project.build.outputTimestamp</property>
+                  <message>The property "project.build.outputTimestamp" must be set on the reactor's root pom.xml to make the build reproducible. Further information at "https://maven.apache.org/guides/mini/guide-reproducible-builds.html".</message>
+                  <!-- https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/ -->
+                  <regex>^([\+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([\.,]\d+(?!:))?)?(\17[0-5]\d([\.,]\d+)?)?([zZ]|([\+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)?$</regex>
```
michael-o, line 339: This does not cover because also a seconds since epoch are allowed. Processing will happily fail when the input format is wrong: https://maven.apache.org/shared/maven-archiver/xref/org/apache/maven/archiver/MavenArchiver.html#L775. I would go with the require only.

michael-o (changes requested):

pom.xml
kwin: Done in 45d63d2.

michael-o (approved):

michael-o:
Althouth this looks technically correct, I do think that this will not work. You have moved the property to a comment and added an enforcer rule, but with the next release the property will be added again and inherited.

kwin:
IMHO maven-release-plugin will only update an existing property but never add it initially.

michael-o:
You are right: https://github.com/apache/maven-release/blob/44ff401499713c981732f3379de2cf6693e0fdef/maven-release-manager/src/main/java/org/apache/maven/shared/release/phase/AbstractRewritePomsPhase.java#L350-L355

kwin:
Seems we need to disable the enforcer check for the build of asf parent itself

michael-o:
What would be your proposal then?

kwin:
Activate the rule only in a profile which is activated only in case a marker file with name `asf-parent.marker` is missing. That would require adding such a file (could be empty) to the local directory.
The only other option is to deal with explicitly given profiles or system properties in case asf-parent is built.

michael-o:
I think we can make the file even hidden because this is a technical detail. Please raise a followup PR for this.

kwin:
I raised a follow-up PR in https://github.com/apache/maven-apache-parent/pull/33. @michael-o Please have a look.

PR #35, opened 2021-02-21
[MPOM-261] create buildinfo file for reproducible builds

kwin:


hboutemy:
we don't need to record buildinfo: see https://github.com/jvm-repo-rebuild/reproducible-central that was done without saving buildinfo at initial build time

artifact:buildinfo is there to check reproducible build on rebuilds

kwin:
@hboutemy Don't you think that buildinfo files should be available by default from Maven Central?
Which parts of Maven Central currently has buildinfo being generated? I only see 288 releases....

hboutemy:
currently, I don't see any value in having buildinfo in Maven Central
And given the format is not really stable, if people publish buildinfo, anybody wanting to consume it would have to hard-code tweaks bsaed on which flavour has been generated

on 288 releases that you see on Reproducible Central, only a few have buildinfo on Central (that's SBT projects...): no Maven project at all has buildinfo

kwin:
@hboutemy Thanks for the answers. I understand that it is too early right now for buildinfo to be published (due to the format not finalized) to Maven Central. Are you also implying that the buildinfo is not necessary even in the long-term for Maven based projects as the relevant information can be derived from pom.xml and MANIFEST.MF or do you agree that the buildinfo in the long term should be published along with the artifacts?
For me the primary goal is to verify that the build artifacts published/downloaded from Central are really based on a specific source. For that the buildinfo is crucial as otherwise you have to rely on heuristics (https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=74682318#Reproducible/VerifiableBuilds-Rebuilding).

hboutemy:
> Are you also implying that the buildinfo is not necessary even in the long-term for Maven based projects as the relevant information can be derived from pom.xml and MANIFEST.MF or do you agree that the buildinfo in the long term should be published along with the artifacts?

in short, currently, even in the long term, I don't see much benefit in publishing buildinfo.

> For me the primary goal is to verify that the build artifacts published/downloaded from Central are really based on a specific source. For that the buildinfo is crucial as otherwise you have to rely on heuristics

We're ok on the goal. But on buildinfo being crucial, I thought when I wrote the spec, but I'm less convinced nowadays after 1 year of experience of really rebuilding and checking if expected reproducible build is really reproducible or not.

Key reason is: *having a buildinfo does not prove that the build is reproducible*

Did you try to rebuild a release that you did not produce yourself at first?

Notice: please continue the discussion, it's useful to dig as we are doing, because we're doing the journey to effective reproducibility check = what we need

hboutemy:
it would be really useful to have some new eyes on reproducing builds
I propose you 1 or 2 very simple projects to rebuild and check that you can get the same result as the reference build: see the beginning of https://github.com/jvm-repo-rebuild/reproducible-central/issues/42
Given the existing content, testing something like maven-scripting-plugin would be really useful

Are you interested to test?

hboutemy:
I'll close this issue as "won't fix"
and probably update official Reproducible Builds site with learning from these years of work: making this buildinfo file so prominent looked the right way when I started, but now, I see that it's not the case

jira-importer:
Resolve #338

PR #40, opened 2021-05-03
[MPOM-244] upload SHA-512 only for source-release to staging repo

kwin:
The SHA-512 checksum should be attached similar to the source-release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -441,31 +441,26 @@ under the License.
           <plugin>
             <groupId>net.nicoulaj.maven.plugins</groupId>
             <artifactId>checksum-maven-plugin</artifactId>
-            <version>1.9</version>
+            <version>1.10</version>
             <executions>
               <execution>
                 <id>source-release-checksum</id>
                 <goals>
-                  <goal>files</goal>
+                  <goal>artifacts</goal>
                 </goals>
+                <configuration>
+                  <algorithms>
+                    <algorithm>SHA-512</algorithm>
+                  </algorithms>
+                  <!-- https://maven.apache.org/apache-resource-bundles/#source-release-assembly-descriptor -->
+                  <includeClassifiers>source-release,source-release-tar</includeClassifiers>
+                  <csvSummary>false</csvSummary>
+                  <!-- attach SHA-512 checksum as well to upload to Maven Staging Repo, 
+                       as this eases uploading from stage to dist and doesn't do harm in Maven Central -->
+                  <attachChecksums>true</attachChecksums>
+                </configuration>
               </execution>
             </executions>
-            <configuration>
-              <algorithms>
-                <algorithm>SHA-512</algorithm>
-              </algorithms>
-              <csvSummary>false</csvSummary>
-              <fileSets>
-                <fileSet>
-                  <directory>${project.build.directory}</directory>
-                  <includes>
-                    <include>${project.artifactId}-${project.version}-source-release.zip</include>
-                    <include>${project.artifactId}-${project.version}-source-release.tar*</include>
-                  </includes>
-                </fileSet>
-              </fileSets>
-              <failIfNoFiles>false</failIfNoFiles><!-- usually, no file to do checksum: don't consider error -->
-            </configuration>
           </plugin>
         </plugins>
       </build>
--- a/pom.xml
+++ b/pom.xml
@@ -453,7 +453,7 @@ under the License.
                     <algorithm>SHA-512</algorithm>
                   </algorithms>
                   <!-- https://maven.apache.org/apache-resource-bundles/#source-release-assembly-descriptor -->
-                  <includeClassifiers>source-release,source-release-tar</includeClassifiers>
+                  <includeClassifiers>source-release</includeClassifiers>
                   <csvSummary>false</csvSummary>
                   <!-- attach SHA-512 checksum as well to upload to Maven Staging Repo, 
                        as this eases uploading from stage to dist and doesn't do harm in Maven Central -->
--- a/pom.xml
+++ b/pom.xml
@@ -462,6 +462,7 @@ under the License.
                 </goals>
                 <configuration>
                   <!-- don't sign SHA-512 checksums (in addition to default MD5 and SHA1) -->
+                  <!-- This does not work due to https://issues.apache.org/jira/browse/MGPG-66 -->
                   <excludes>
                     <exclude>**/*.sha512</exclude>
                   </excludes>
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@ under the License.
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-gpg-plugin</artifactId>
-          <version>1.6</version>
+          <version>3.0.1</version>
           <configuration>
             <gpgArguments>
               <arg>--digest-algo=SHA512</arg>
@@ -450,7 +450,7 @@ under the License.
               </execution>
             </executions>
           </plugin>
-          <!-- We want to sign the artifact, the POM, and all attached artifacts -->
+          <!-- We want to sign the artifact, the POM, and all attached artifacts (except for SHA-512 checksum) -->
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-gpg-plugin</artifactId>
@@ -460,13 +460,6 @@ under the License.
                 <goals>
                   <goal>sign</goal>
                 </goals>
-                <configuration>
-                  <!-- don't sign SHA-512 checksums (in addition to default MD5 and SHA1) -->
-                  <!-- This does not work due to https://issues.apache.org/jira/browse/MGPG-66 -->
-                  <excludes>
-                    <exclude>**/*.sha512</exclude>
-                  </excludes>
-                </configuration>
               </execution>
             </executions>
           </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -435,6 +435,8 @@ under the License.
                 <goals>
                   <goal>artifacts</goal>
                 </goals>
+                <!-- execute prior to maven-gpg-plugin:sign due to https://github.com/nicoulaj/checksum-maven-plugin/issues/112 -->
+                <phase>post-integration-test</phase>
                 <configuration>
                   <algorithms>
                     <algorithm>SHA-512</algorithm>
--- a/pom.xml
+++ b/pom.xml
@@ -424,19 +424,6 @@ under the License.
               </execution>
             </executions>
           </plugin>
-          <!-- We want to sign the artifact, the POM, and all attached artifacts -->
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-gpg-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>sign-release-artifacts</id>
-                <goals>
-                  <goal>sign</goal>
-                </goals>
-              </execution>
-            </executions>
-          </plugin>
           <!-- calculate checksums of source release for Apache dist area -->
           <plugin>
             <groupId>net.nicoulaj.maven.plugins</groupId>
@@ -454,10 +441,30 @@ under the License.
                   </algorithms>
                   <!-- https://maven.apache.org/apache-resource-bundles/#source-release-assembly-descriptor -->
                   <includeClassifiers>source-release</includeClassifiers>
+                  <excludeMainArtifact>true</excludeMainArtifact>
                   <csvSummary>false</csvSummary>
                   <!-- attach SHA-512 checksum as well to upload to Maven Staging Repo, 
                        as this eases uploading from stage to dist and doesn't do harm in Maven Central -->
                   <attachChecksums>true</attachChecksums>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <!-- We want to sign the artifact, the POM, and all attached artifacts -->
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>sign-release-artifacts</id>
+                <goals>
+                  <goal>sign</goal>
+                </goals>
+                <configuration>
+                  <!-- don't sign SHA-512 checksums (in addition to default MD5 and SHA1) -->
+                  <excludes>
+                    <exclude>**/*.sha512</exclude>
+                  </excludes>
                 </configuration>
               </execution>
             </executions>


pom.xml
```diff
@@ -441,31 +441,26 @@ under the License.
...
                 <goals>
-                  <goal>files</goal>
+                  <goal>artifacts</goal>
                 </goals>
+                <configuration>
+                  <algorithms>
+                    <algorithm>SHA-512</algorithm>
+                  </algorithms>
+                  <!-- https://maven.apache.org/apache-resource-bundles/#source-release-assembly-descriptor -->
+                  <includeClassifiers>source-release,source-release-tar</includeClassifiers>
```
michael-o, line 456: Where did you get the `source-release-tar` classifier from?

kwin: My bad, I think all descriptors use the same classifier (https://github.com/apache/maven-resources/blob/e9eaa1baebb72795555eba015eac9b40f288e931/apache-source-release-assembly-descriptor/src/main/assemblies/source-release.xml#L22)

michael-o (changes requested):
I consider this change non-sense and incomplete:
* ASF releases are solely distributed through dist.apache.org
* Generating checksums on the fly for remote repos is done by Maven Resolver
* Uploading checksums for a subset of items makes no sense

Note that Maven Resolver uses checksums to detect transport corruption only. Verification *must* happen with signatures.

michael-o:
@hboutemy

kwin:
> ASF releases are solely distributed through dist.apache.org

Not in all Apache projects, look at https://repo1.maven.org/maven2/org/apache/sling/org.apache.sling.api/2.9.0/ or https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.scr/2.1.8/. Maven Central has the source-release artifacts as well.

> Generating checksums on the fly for remote repos is done by Maven Resolver

Only MD5 and SHA1, SHA512 has been explicitly disabled for performance reasons (https://issues.apache.org/jira/browse/MRESOLVER-140)

> Uploading checksums for a subset of items makes no sense

ASF policy enforces that only for those releases. For the binaries it is never checked (because as you said, Maven will only use MD5/SHA1 to detect transport corruption)

michael-o:
> 
> 
> > ASF releases are solely distributed through dist.apache.org
> 
> Not in all Apache projects, look at https://repo1.maven.org/maven2/org/apache/sling/org.apache.sling.api/2.9.0/ or https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.scr/2.1.8/. Maven Central has the source-release artifacts as well.

They must comply with: https://infra.apache.org/release-distribution.html. They can provide more if they like, but not less than that.

> > Generating checksums on the fly for remote repos is done by Maven Resolver
> 
> Only MD5 and SHA1. SHA512 has been explicitly disabled for performance reasons (https://issues.apache.org/jira/browse/MRESOLVER-140)

What holds you off to enable this?

> > Uploading checksums for a subset of items makes no sense
> 
> ASF policy enforces that only for those releases (https://infra.apache.org/release-distribution.html#sigs-and-sums). For the binaries it is never checked (because as you said, Maven will only use MD5/SHA1 to detect transport corruption) and ASF doesn't care about the binaries anyhow.

ASF never provides binary. The only official release is a source tarball/zip. Binaries are courtesy. So if you want to point someone to an Apache release, point to downloads.apache.org. That's it.

kwin:
To be honest, it is a bit tough to discuss with you. Also language like "non-sense" does not really help. But let me try to respond to your complaints:

> They must comply with: https://infra.apache.org/release-distribution.html. They can provide more if they like, but not less than that.

Yes, even Apache Sling and Felix do that, but this is IMHO not relevant in the context of this PR.

> What holds you off to enable this?

This is only possible with command line options and not via pom.xml configuration. To distribute that among a team is much more effort than on relying on a plugin available from Maven Central which is even compatible with Apache Maven prior 3.8.1 (which is IMHO the first release shipping with Maven Resolver 1.6.x supporting SHA-512 on demand)

> ASF never provides binary. The only official release is a source tarball/zip. Binaries are courtesy. So if you want to point someone to an Apache release, point to downloads.apache.org. That's it.

Same as above. IMHO not really relevant in the context of this PR. In fact a lot of ASF projects provide binaries via Maven Central although ASF only mandates distribution of source archives. The whole point about this PR is being able to use the Staging Repo as the single source for dist and Maven Central (as that eases voting, validating a release candidate).

michael-o:
Does [this](https://maven.apache.org/maven-release/maven-release-plugin/perform-mojo.html#arguments) work for you? I also believe that if you attach checksums Resolver will calculate checksums of checksums.

kwin:
I decided against using maven-resolver-impl for calculating SHA-512 for the following reasons:
1. It requires Resolver 1.6+ which only ships with Maven 3.8.1
2. It will calculate SHA-512 for all artifacts instead only the source-release ones which is IMHO too much overhead and not really useful

Although the SHA-512 artifact being calculated by the checksum-maven-plugin will get another MD5/SHA1 checksum when being deployed that IMHO doesn't do any harm.

kwin:
@hboutemy I would appreciate your input on this as well.

hboutemy:
IIUC, in the end, you end-up about publishing sha512 checksum to Central **only for source-release**?
I must admit that this will ease people's filing Apache dist area, because copying from the build workspace is cumbersome, and I don't expect usual release manager thinks about it after doing 'mvn release:perform', then he has hard time to fill Apache dist area and recalculate the sha512 once the vote is done
I like this limited scope sha512 idea: @michael-o don't you struggle to get the sha512 when committing source-release to Apache dista area?

on the implementation side, this means that the current commit is big because there was POM reformatting (change of order of 2 plugins), but the real update is about adding `<attachChecksums>true</attachChecksums>` to checksum plugin for source-release: nice
question: is the reordering necessary?

kwin:
The reordering is no longer necessary with m-gpg-p 3.0.1 I will revert

hboutemy:
I understand why you need to exclude asc from checksum, if asc was produced before checksum
I don't understand why we need to ensure that asc was produced before checksum: if by chance it was not produced yet, let benefit from chance, isn't it?

kwin:
Let me summarize the status quo (without PR):

1. `gpg-maven-plugin:sign` executed in phase `verify`, calculates ASC files for almost all attached files (except for MD5, SHA1 and ASC)
1. `checksum-maven-plugin:files` executed in phase `verify`, calculates the checksum based on filename (considering the extension) but does not attach it

With this PR the order is inverted:

1. `checksum-maven-plugin:artifacts` executed in phase `verify`, calculates the checksum based on classifier (disregarding the extension) and attaching it to project
1. `gpg-maven-plugin:sign` executed in phase `verify`, calculates ASC files for almost all attached files (except for MD5, SHA1, SHA-512 and ASC)

Without the inverted order the `checksum-maven-plugin` would calculate a checksum for the `ASC` for  `source-release` as well, which is certainly not desired.

hboutemy:
ah, ok, I didn't get that attaching checksums required to switch from `files` goal to `artifacts`, and no exclusion based on extension exists yet...
thanks for the explanation
@michael-o do you get the idea of publishing `source-release` sha512 to remote repository to ease filling Apache distribution?

hboutemy:
@kwin wouldn't `files` goal https://checksum-maven-plugin.nicoulaj.net/files-mojo.html be an option? It supports attach, and would not suffer from the ordering prerequisite (which is not something that is guaranteed in a phase

kwin:
files goal does never attach with a classifier, therefore we can't use it

hboutemy:
ok, then it's safer to bind the `checksum-maven-plugin:artifacts` goal to `post-integration-test` phase

kwin:
I can do that but for clarity reason I would still reorder in the pin

michael-o (commented):
I have checked the change. Two questions:
* Will Resolver generate a checksum of the checksum when the checkusm is attached?
* What will happen if Resolver is configured to generate SHA-512 sums?

kwin:
I answered the first question already in https://github.com/apache/maven-apache-parent/pull/40#issuecomment-831703133:

> Although the SHA-512 artifact being calculated by the checksum-maven-plugin will get another MD5/SHA1 checksum when being deployed that IMHO doesn't do any harm.

For the 2nd question, I have not tried, but most probably resolver will just overwrite the existing file, for now this isn't an issue, because I don't see a reason why ASF projects should generate SHA-512 by Maven Resolver, it won't ever be enabled by default for performance reasons.

michael-o:
> 
> 
> I answered the first question already in [#40 (comment)](https://github.com/apache/maven-apache-parent/pull/40#issuecomment-831703133):
> 
> > Although the SHA-512 artifact being calculated by the checksum-maven-plugin will get another MD5/SHA1 checksum when being deployed that IMHO doesn't do any harm.
> 
> For the 2nd question, I have not tried, but most probably resolver will just overwrite the existing file (with the same content), for now this isn't an issue, because I don't see a reason why ASF projects should generate SHA-512 by Maven Resolver, it won't ever be enabled by default for performance reasons.

Even if Resolver overwrites, Nexus Staging might reject this because the file is uploaded twice. You should check that.

kwin:
I think this should be handled in the resolver in https://github.com/apache/maven-resolver/blob/17ea285d40e8cdd183d45b074f74aae6d9f3d3fc/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnector.java#L286.

michael-o:
> 
> 
> I think this should be handled in the resolver in https://github.com/apache/maven-resolver/blob/17ea285d40e8cdd183d45b074f74aae6d9f3d3fc/maven-resolver-connector-basic/src/main/java/org/eclipse/aether/connector/basic/BasicRepositoryConnector.java#L286.

Why? Then you need to add excludes and transport through layers...

kwin:
I rather meant handled by throwing an exception if the checksum artifact name collides with one of the attached artifacts.

hboutemy:
FYI, I tried `mvn -Papache-release deploy` with the sha512 attachments and got https://repository.apache.org/content/repositories/snapshots/org/apache/apache/24-SNAPSHOT/

michael-o:
@hboutemy I wonder what will happen during a release: Maven Resolver uploads SHA-512, this change uploads it. Will Nexus staging reject it? This should be tested before merged.

hboutemy:
the day Maven Resolver will be configured to upload sha512, we'll remove this hack (because, yes, it's a hack to ease our release process)
When do you expect Maven Resolver to publish sha512 in addition to sha1 and md5 in real world?

michael-o:
> 
> 
> the day Maven Resolver will be configured to upload sha512, we'll remove this hack (because, yes, it's a hack to ease our release process)
> When do you expect Maven Resolver to publish sha512 in addition to sha1 and md5 in real world?

People can already configure it. I just want to avoid surprises for all Apache folks. E.g., Mark Thomas does this actively now.

michael-o:
See here: https://repo1.maven.org/maven2/org/apache/tomcat/tomcat-catalina/10.0.6/

Produced by Resoler via Ant.

hboutemy:
FYI, I added some instructions to calculate sha512 in dist: https://github.com/apache/maven-site/commit/29bc21e3dc9d2bb7ff61ce49cecf8a8600b9b5d1
and while at it, I recalculated checksums for all our latest releases and found 6 fingerprints not exactly using the expected format (fixed in revision 48099): this shows concrete example how people struggle to add the sha512 to dist area...

hboutemy:
> People can already configure it. I just want to avoid surprises for all Apache folks. E.g., Mark Thomas does this actively now.

can you show me, please? I don't know how to do that

notice: I mean using Maven, because that's the Maven ASF parent POM that we're tweaking...

michael-o:
> 
> 
> FYI, I added some instructions to calculate sha512 in dist: [apache/maven-site@29bc21e](https://github.com/apache/maven-site/commit/29bc21e3dc9d2bb7ff61ce49cecf8a8600b9b5d1)
> and while at it, I recalculated checksums for all our latest releases and found 6 fingerprints not exactly using the expected format (fixed in revision 48099): this shows concrete example how people struggle to add the sha512 to dist area...

I have done this massively in Core ITs: https://github.com/apache/maven-integration-testing/commit/672728a363c8b3503250589504efbfbea152e7e9

michael-o:
> 
> 
> > People can already configure it. I just want to avoid surprises for all Apache folks. E.g., Mark Thomas does this actively now.
> 
> can you show me, please? I don't know how to do that
> 
> notice: I mean using Maven, because that's the Maven ASF parent POM that we're tweaking...

See https://maven.apache.org/resolver/configuration.html for `aether.checksums.algorithms=SHA-512`. Resolver should generate and update them. Please try in your setup.

hboutemy:
oh nice, I didn't know that content!
I ran `mvn -Papache-release deploy -Daether.checksums.algorithms="SHA-512,SHA-1,MD5"` and got id 86 of https://repository.apache.org/content/repositories/snapshots/org/apache/apache/24-SNAPSHOT/
no problem for now

hboutemy:
changed pom.xml version to 23 and ran the same `mvn -Papache-release deploy -Daether.checksums.algorithms="SHA-512,SHA-1,MD5"`
=> perfect staging repository: https://repository.apache.org/content/repositories/orgapacheapache-1020/org/apache/apache/23/

michael-o:
> changed pom.xml version to 23 and ran the same `mvn -Papache-release deploy -Daether.checksums.algorithms="SHA-512,SHA-1,MD5"`
> => perfect staging repository: https://repository.apache.org/content/repositories/orgapacheapache-1020/org/apache/apache/23/

That is nice. One can manually delete the redundant files from staging.

michael-o (approved):

hboutemy:
I did not test yet, but why don't we just add `<aether.checksums.algorithms>SHA-512,SHA-1,MD5</aether.checksums.algorithms>` property to the parent POM?
wouldn't it activate sha512 publishing by Maven for absolutely every artifact?

kwin:
@hboutemy See my answer in https://github.com/apache/maven-apache-parent/pull/40#issuecomment-831703133

hboutemy:
@kwin ok, thanks for the link
let's merge this

michael-o:
@kwin Thanks for bearing with me. I am very picky when it affects all ASF committers.

hboutemy:
reading your comment, it makes me feel that for a very long time, we're probably conflating:
- hash format for Maven repository format (= hash only, defined by Maven repository format and Maven code that reads/writes it)
- hash format for Apache distribution

kwin:
ASF mandates this format: https://www.apache.org/info/verification.html

khmarbaise:
The referenced page of [apache explicitly mentions](https://www.apache.org/info/verification.html) the usage of `shasum/md5sum/certUtil` tool which I have mentioned as well in [MNG-6784](https://issues.apache.org/jira/browse/MNG-6784). 

The format from my point of view is a consequence by using the appropriate tools... if not otherwise explicitly mentioned I would assume that the format is the one which can be consumed by the given tools. This is currently not the case.

jira-importer:
Resolve #456

PR #48, opened 2021-06-17
[MPOM-264] Parameterize maven-compiler-plugin with parameter "release"

kwin:
when running on JDK 9+

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -370,6 +370,44 @@ under the License.
   </build>
 
   <profiles>
+    <profile>
+      <id>jdk9-or-higher</id>
+      <activation>
+        <!-- Syntax according to http://maven.apache.org/enforcer/enforcer-rules/versionRanges.html -->
+        <jdk>[9,)</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <!-- Set property "maven.compiler.release" for m-compiler-p with 
+          https://www.mojohaus.org/build-helper-maven-plugin/regex-property-mojo.html.
+          "maven.compiler.release" sets API classpath, source and target, see 
+          https://docs.oracle.com/javase/9/tools/javac.htm#GUID-AEEC9F07-CB49-4E96-8BC7-BCC2C7F725C9__GUID-D343F6B4-3FDD-43A8-AD24-43DD70214471.
+          This takes precedence over "maven.compile.source" and "maven.compiler.target"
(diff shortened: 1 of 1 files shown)


rfscholte:
I think this is too dangerous. What if projects are using toolchains?

pom.xml
```diff
@@ -370,6 +370,44 @@ under the License.
...
+            <executions>
+              <execution>
+                <id>set-compiler-release-parameter</id>
+                <goals>
+                  <goal>regex-property</goal>
+                </goals>
+                <configuration>
+                  <name>maven.compiler.release</name>
+                  <!-- strip leading "1." from maven.compiler.target -->
+                  <regex>1\.(.*)</regex>
```
michael-o, line 400: What if the target contains `11`? This is too much magic here...

kwin: If the value doesn't start with "1." it is just used as is.

kwin: I am open for other proposals on how to achieve this while keeping backwards compatibility and only being forced to define the target java version in one property.

michael-o: Thanks for the information.

kwin:
> I think this is too dangerous. What if projects are using toolchains?

Is it possible to add a profile condition for toolchains?

rfscholte:
No, the JDK activation only works for the Maven runtime.

kwin:
@rfscholte Any proposal how to put this functionality into ASF parent pom or do you rather think every project should configure "release" on a project level.

rfscholte:
This is a returning discussion: if we need to update source/target(/release) it means there is just no sane default value. IMO projects should always define these values themselves to prevent surprises by updating a parent.

kwin:
This PR is not about defining the default value but about setting the target java version in one(!) property and leverage it both in Java < 9 and 9 + optimally

rfscholte:
Toolchain will still be an issue.

rfscholte:
Another argument: if both Michael and I don't see it right away what's happening here, it is "magic" that should happen inside a plugin (where the magic can be tested). It will make the pom too hard to maintain (which would bring us back to the Ant buildscript-master: only one person will understand what is happening here, there rest either ignores it or doesn't dare to touch it)

kwin:
I opened https://github.com/codehaus-plexus/plexus-compiler/issues/140 to allow to set the parameter "maven.compiler.release" without risking to break the build for JDK < 9.

jira-importer:
Resolve #479

PR #61, opened 2022-01-13
[MPOM-286] org.apache:apache should conform with needed Sonatype checksums on deployment

Tibor17:


Tibor17:
@slawekjaranowski
Here is the improvement. Maybe we should ping the team.

michael-o (commented):
Why? Isn't it enough to have it in the release profile only?

Tibor17:
@michael-o
I changed to checksum-maven-plugin. Not the GPG.

slawekjaranowski (changes requested):
Don't do it in this way. `gpg-plugin` must be only in `apache-release` profile otherwise we will have problem with all build. 

Probably gpg is only configured on local machines, and I suppose that only for more advanced users, usually is not configured, so many contributors may have such problem like in GH checks for this PR.

This change don't resolve your problem.

Tibor17:
@slawekjaranowski 
Check the commit again, I am moving only `checksum-maven-plugin`, not the GPG.

michael-o:
I still don't understand the purpose of this for non-releases.

slawekjaranowski:
I afraid that sha-512 is not required for Sonatype ...

it only create signature for artifact with classifier `source-release` ... but source-release is created in apache-release profile.

kwin:
What issue are you trying to solve here? Neither ASFs Nexus nor Maven Central require SHA-512 checksums

hboutemy:
so really, no, don't do this, because:
1. the need is not clear (not to say wrong, which is what I think currently)
2. the implementation is wrong: you're trying to calculate a checksum for an artifact that is created only in apache-release profile

if you want to use apache-release profile with a SNAPSHOT, don't hesitate to do it: it perfectly works

what is currently defined in apache-release profile is consistent: please don't break this consistency

slawekjaranowski:
@Tibor17 - can we closed it?

jira-importer:
Resolve #498

PR #72, opened 2022-03-13
Stage site build (that uses site-pom.xml)

slawekjaranowski:
Corresponding to the latest change in shared action.

https://github.com/apache/maven-gh-actions-shared/pull/40

Site build output is attached to action build, can be downloaded and check result.


slawekjaranowski:
next build will fail without it

slachiewicz (approved):

PR #111, opened 2022-10-10
[MPOM-347] Allow custom Release Distribution Repository

pan3793:
Same as snapshots, this makes developers deploy a release version to a custom repo easily.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,8 +69,8 @@ under the License.
   <distributionManagement>
     <repository>
       <id>apache.releases.https</id>
-      <name>Apache Release Distribution Repository</name>
-      <url>https://repository.apache.org/service/local/staging/deploy/maven2</url>
+      <name>${distMgmtReleasesName}</name>
+      <url>${distMgmtReleasesUrl}</url>
     </repository>
     <snapshotRepository>
       <id>apache.snapshots.https</id>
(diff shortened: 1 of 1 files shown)


michael-o:
Why?

pan3793:
why does the snapshot repo allow custom but the release does not?

michael-o:
Good question...

pan3793:
one use case: some users/developers may want to have a quick patched version to deploy to private repo based on a released version, e.g. someone wants a v1.2.3-1 w/ some internal changes based on v1.2.3, w/ this patch, they can simply run 
```
mvn versions:set -DnewVersion=1.2.3-1 -DgenerateBackupPoms=false
mvn deploy -DdistMgmtReleasesName=<private-repo-name> -DdistMgmtReleasesUrl=<private-repo-url>
```
instead of touching `pom.xml`

olamy (approved):
even if you can override it in a child (see how it's done in Archiva pom https://github.com/apache/archiva-parent/blob/9f51d7eb3427bca02b7f7df6ffe02e960d8c7a69/pom.xml#L46 )

this seems to be a legitimate option.

pan3793:
Yes, @olamy we have done the same thing in Apache Kyuubi (Incubating) https://github.com/apache/incubator-kyuubi/blob/2cb34c4d007ac19b359767963225ee422d765618/pom.xml#L281-L292

pan3793:
I use "releases" for consistency as we already used "snapshots", are you inclined towards "release" or "releases"?

jira-importer:
Resolve #444

PR #155, opened 2023-05-23
[MPOM-289] Use properties to define the versions of plugins

slawekjaranowski:
discussion invite sent to release-discuss@apache.org: https://lists.apache.org/list?release-discuss@apache.org:2023-6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,6 +96,34 @@ under the License.
     <surefire.version>3.1.0</surefire.version><!-- for surefire, failsafe and surefire-report -->
     <maven.plugin.tools.version>3.9.0</maven.plugin.tools.version><!-- for m-plugin-p and maven-plugin-annotations -->
     <assembly.tarLongFileMode>posix</assembly.tarLongFileMode>
+
+    <!-- plugins versions -->
+    <apacheRatPluginVersion>0.15</apacheRatPluginVersion>
+    <mavenAntrunPluginVersion>3.1.0</mavenAntrunPluginVersion>
+    <mavenAssemblyPluginVersion>3.6.0</mavenAssemblyPluginVersion>
+    <mavenCleanPluginVersion>3.2.0</mavenCleanPluginVersion>
+    <mavenCompilerPluginVersion>3.11.0</mavenCompilerPluginVersion>
+    <mavenDependencyPluginVersion>3.6.0</mavenDependencyPluginVersion>
+    <mavenDeployPluginVersion>3.1.1</mavenDeployPluginVersion>
+    <mavenEarPluginVersion>3.3.0</mavenEarPluginVersion>
+    <mavenEnforcerPluginVersion>3.3.0</mavenEnforcerPluginVersion>
+    <mavenGpgPluginVersion>3.1.0</mavenGpgPluginVersion>
+    <mavenHelpPluginVersion>3.4.0</mavenHelpPluginVersion>
+    <mavenInstallPluginVersion>3.1.1</mavenInstallPluginVersion>
+    <mavenInvokerPluginVersion>3.5.1</mavenInvokerPluginVersion>
+    <mavenJarPluginVersion>3.3.0</mavenJarPluginVersion>
+    <mavenJavadocPluginVersion>3.5.0</mavenJavadocPluginVersion>
+    <mavenProjectInfoReportsPluginVersion>3.4.3</mavenProjectInfoReportsPluginVersion>
+    <mavenReleasePluginVersion>3.0.0</mavenReleasePluginVersion>
+    <mavenRemoteResourcesPluginVersion>3.1.0</mavenRemoteResourcesPluginVersion>
+    <mavenResourcesPluginVersion>3.3.1</mavenResourcesPluginVersion>
+    <mavenScmPluginVersion>2.0.1</mavenScmPluginVersion>
+    <mavenScmPublishPluginVersion>3.2.1</mavenScmPublishPluginVersion>
+    <mavenShadePluginVersion>3.4.1</mavenShadePluginVersion>
+    <mavenSitePluginVersion>3.12.1</mavenSitePluginVersion>
+    <mavenSourcePluginVersion>3.3.0</mavenSourcePluginVersion>
+    <mavenWarPluginVersion>3.3.2</mavenWarPluginVersion>
+
     <project.build.outputTimestamp>2022-12-11T19:18:25Z</project.build.outputTimestamp>
   </properties>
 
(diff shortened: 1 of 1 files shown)


slachiewicz (approved):

pom.xml
```diff
@@ -96,6 +96,34 @@ under the License.
...
+    <mavenProjectInfoReportsPluginVersion>3.4.3</mavenProjectInfoReportsPluginVersion>
+    <mavenReleasePluginVersion>3.0.0</mavenReleasePluginVersion>
+    <mavenRemoteResourcesPluginVersion>3.1.0</mavenRemoteResourcesPluginVersion>
+    <mavenResourcesPluginVersion>3.3.1</mavenResourcesPluginVersion>
+    <mavenScmPluginVersion>2.0.1</mavenScmPluginVersion>
+    <mavenScmPublishPluginVersion>3.2.1</mavenScmPublishPluginVersion>
+    <mavenShadePluginVersion>3.4.1</mavenShadePluginVersion>
+    <mavenSitePluginVersion>3.12.1</mavenSitePluginVersion>
+    <mavenSourcePluginVersion>3.3.0</mavenSourcePluginVersion>
+    <mavenWarPluginVersion>3.3.2</mavenWarPluginVersion>
```
ctubbsii, line 125: The idea is nice, but can we use a naming convention that is more sortable, and groups version-related properties apart from other properties, and uses the artifact name as part of the property, so it's easy to grep for these? Something like `maven-release-plugin -> version.maven-release-plugin`.

olamy:
I like the idea. Then it will be possible to use interpolation in it pom files to use same version and avoid some download.

pom.xml
slawekjaranowski: I want to reuse those values in many place also in documentation in .vm files ... but we can not use 'dot' for velocity templates.

ctubbsii: I'm not sure I understand your use case, but couldn't you use the maven-resources-plugin to filter Maven variables into your velocity templates? Or use your own velocity-compatible variables?

A slight variation that might accommodate that use case would be to use another dash, rather than a dot. The important principle is that "version" comes first, so when the properties are all sorted, the ones that specify versions are all grouped together, and that the part after "version" is the same as the artifact name, so it's easy to grep.

ctubbsii: @slawekjaranowski The [maven-site-plugin documentation](https://maven.apache.org/plugins/maven-site-plugin/examples/creating-content.html#Filtering) says you can use `$context.get("my.property")`, so that's not a problem.

slawekjaranowski: @ctubbsii I know such workarounds for sitedocs ... but I would like to use properties as simply as possible without additional hacks.
So I will merge as is - if no more objections.

ctubbsii: I stand by my objection that I think the naming convention could be better... even if you choose not to use dots, using the same name as the plugin artifact and using the word "version" as the prefix rather than the suffix, is better for quickly finding these. It wasn't a problem when there were only a few, but as you're making a bunch of them, keeping them organized becomes more important.

I still think the workaround for the dot in the velocity templates is not a problem... I don't think this should cater to that edge case, because dots are common in Maven and Java, even if velocity doesn't like them. But, even if you do cater to that edge case, there's plenty of ways to improve the naming scheme without using dots.

rmannibucau: Agree dots are more expected and velocity usage can be enhanced to support the lookup. In any case camel case looks like inconsistent with maven expressions so maybe not?

gnodet (approved):
Looks good, I don't have any particular opinion about the naming...

michael-o (approved):
Very nice!

pom.xml
ctubbsii: > I respect your opinion, as you look at linked issues such convention is already used in many Maven projects
> 
> I also was trying to find what kind of properties are used in other ASF project ... I have found, eg: `<artifactId>.version` `<artifactId>-version` `<artifactId with dot instead of dash>.version`
> 
> I didn't found properties with versions word as prefix.

I haven't found many that do this either... but in my experience, it's a much better convention to put version at the front, especially when using plugins like sortpom-maven-plugin to organize the POM and sort the properties. It's very convenient to have all the versions grouped together. Many projects don't know about sortpom-maven-plugin, and don't use it, though, despite it being a superb plugin for POM quality control.

> 
> I afraid that there is no convention which can be ok for all, how many projects, so many opinions

Agreed. But my hope is that the reasons I provided are worth serious consideration, and more people will adopt similar conventions on the same grounds, in the absence of arguments against. Specifically: sort-ability and search-ability are the reasons I gave (I think both help with maintainability: smaller diffs, easier to find and update things).

ctubbsii:
Oh, one other comment on this: I've noticed that overriding versions in the child POMs by overriding the property does not cause versions-maven-plugin to notice newer versions when running: `mvn -U versions:display-pluginupdates`. That's not really an issue for this project... more a deficiency in versions-maven-plugin... but I wanted to mention it here, because if a user used these new properties to override the plugin versions instead of defining a version override in their own `<pluginManagement>` section, they might not notice that the versions-maven-plugin has that limitation.

hgschmie:
I use `dep.plugin.<plugin-name>.version` properties for a very long time: https://github.com/basepom/basepom/blob/main/poms/foundation/pom.xml#L236-L243

Using a hierarchical naming scheme gives you an opportunity to sort those (e.g. with the sortpom plugin - https://github.com/Ekryd/sortpom) and not have the versions strewn all over the place (the names all align on `dep.plugin` first).

pom.xml
hboutemy: - camelCase for such properties is really unusual to many, and *much harder to read* than having a real separator (dot, underscore, dash); Choosing camelCase only for .vm simplification looks IMHO a bad compromise
- whatever the convention for splitting fields, having version as first field for sortability seems a very good idea
- on plugin field: is there an issue with using direct artifactId?
- remember that this PR introduces new properties that many ASF child projects will later use to override: changing the names later will cause much friction, then the choice has to be well thought

hboutemy:
> Oh, one other comment on this: I've noticed that overriding versions in the child POMs by overriding the property does not cause versions-maven-plugin to notice newer versions when running: `mvn -U versions:display-plugin-updates`. That's not really an issue for this project... more a deficiency in versions-maven-plugin... but I wanted to mention it here, because if a user used these new properties to override the plugin versions instead of defining a version override in their own `<pluginManagement>` section, they might not notice that the versions-maven-plugin has that limitation.

is there a GH issue opened about that?

pom.xml
bmarwell: +1 for dots

slawekjaranowski: Ok, I see that more of us is for using dots, and we have also idea for prefix.

So what do you think about:
`version.<artifactId>` - like: `version.maven-release-plugin` - thanks @ctubbsii 😄

slawekjaranowski:
> Oh, one other comment on this: I've noticed that overriding versions in the child POMs by overriding the property does not cause versions-maven-plugin to notice newer versions when running: `mvn -U versions:display-plugin-updates`. That's not really an issue for this project... more a deficiency in versions-maven-plugin... but I wanted to mention it here, because if a user used these new properties to override the plugin versions instead of defining a version override in their own `<pluginManagement>` section, they might not notice that the versions-maven-plugin has that limitation.

You can use: `mvn versions:display-property-updates` in such case

pom.xml
olamy: perso I usually use `<artifactId>.version` such `<maven-release-plugin.version>` 
it looks more similar to `object.property`

ctubbsii:
> You can use: `mvn versions:display-property-updates` in such case

Thanks @slawekjaranowski ! That's good to know.

pom.xml
martin-g: > perso I usually use `<artifactId>.version` such `<maven-release-plugin.version>`
> it looks more similar to `object.property`

+1 


What about the old properties ?
```
<surefire.version>3.1.0</surefire.version><!-- for surefire, failsafe and surefire-report -->
    <maven.plugin.tools.version>3.9.0</maven.plugin.tools.version>
```

Are you going to change them to be consistent with the new ones ? Any concerns about backward compatibility ?

pom.xml
```diff
@@ -96,6 +96,34 @@ under the License.
     <surefire.version>3.1.0</surefire.version><!-- for surefire, failsafe and surefire-report -->
     <maven.plugin.tools.version>3.9.0</maven.plugin.tools.version><!-- for m-plugin-p and maven-plugin-annotations -->
     <assembly.tarLongFileMode>posix</assembly.tarLongFileMode>
+
+    <!-- plugins versions -->
+    <apacheRatPluginVersion>0.15</apacheRatPluginVersion>
+    <mavenAntrunPluginVersion>3.1.0</mavenAntrunPluginVersion>
+    <mavenAssemblyPluginVersion>3.6.0</mavenAssemblyPluginVersion>
```
bmarwell, line 103: Please stay consistent in Naming: `surefire.version` != `mavenAssemblyPluginVersion`.

If velocity has problems, let's fix velocity. In freemarker we could just use another syntax, I expect velocity to have the same ability. 3rd party tools (ie non-maven) should not have impact on naming.

bmarwell (changes requested):

michael-o:
We can have endless debates about a format, at the end you cannot please everyone. Let's stick to the one we have, it is established and move on.

martin-g:
> Let's stick to the one we have, it is established and move on.

Which one exactly ?
```
    <surefire.version>3.1.0</surefire.version>       ONE
    <maven.plugin.tools.version>3.9.0</maven.plugin.tools.version> TWO
    <apacheRatPluginVersion>0.15</apacheRatPluginVersion> THREE
```

martin-g:
FWIW I vote for `apache-rat-plugin.version`
It is much easier to copy/paste the `artifactId` and append/prepend `version`.

pom.xml
ctubbsii: I would suggest keeping any existing ones, and only following a consistent convention for new ones, for now. Changing the existing ones to follow the same convention (with a possible attempt for backwards compatibility, through the use of property referencing another property, or through a profile activated using the legacy property, or something else), could be done in a subsequent PR, if somebody wanted to do it. For now, it would be enough to just establish the convention using new properties.

ctubbsii:
I created an alternate implementation for my ideas from this discussion at #158, that adopts the `version.$artifactId` naming convention, and addresses the issue with the existing version properties, as well as fixes a few other tiny things (details in the description of that PR).

slawekjaranowski:
suppressed by #158

PR #156, opened 2023-05-23
[MPOM-332] Remove outdated m-deploy-p configuration

slawekjaranowski:
There is no parameter updateReleaseInfo in current plugin version


jira-importer:
Resolve #504

PR #158, opened 2023-06-06
[MPOM-289] Use properties for all plugin versions (alternate)

ctubbsii:
Alternate implementation to PR #155 to demonstrate my alternative
proposal with the naming convention of `version.artifactId` for
properties. The properties are sorted in the properties section.

This also includes individual properties for the plugins that use
`${maven.plugin.tools.version}` and `${surefire.version}`, so they can
be set individually, but will default to using the current properties
that set the version for multiple plugins at once.

This also fixes the M2Eclipse profile, `only-eclipse`, to remove the
upper bound on the version for maven-remote-resources-plugin, because
the version specified in this POM has already exceeded the that upper
bound, and the profile would not have been working correctly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,16 +109,13 @@ under the License.
     <version.maven-deploy-plugin>3.1.1</version.maven-deploy-plugin>
     <version.maven-ear-plugin>3.3.0</version.maven-ear-plugin>
     <version.maven-enforcer-plugin>3.3.0</version.maven-enforcer-plugin>
-    <version.maven-failsafe-plugin>${surefire.version}</version.maven-failsafe-plugin>
     <version.maven-gpg-plugin>3.1.0</version.maven-gpg-plugin>
     <version.maven-help-plugin>3.4.0</version.maven-help-plugin>
     <version.maven-install-plugin>3.1.1</version.maven-install-plugin>
     <version.maven-invoker-plugin>3.5.1</version.maven-invoker-plugin>
     <version.maven-jar-plugin>3.3.0</version.maven-jar-plugin>
     <version.maven-javadoc-plugin>3.5.0</version.maven-javadoc-plugin>
-    <version.maven-plugin-annotations>${maven.plugin.tools.version}</version.maven-plugin-annotations>
-    <version.maven-plugin-plugin>${maven.plugin.tools.version}</version.maven-plugin-plugin>
-    <version.maven-plugin-report-plugin>${maven.plugin.tools.version}</version.maven-plugin-report-plugin>
+    <version.maven-plugin-tools>${maven.plugin.tools.version}</version.maven-plugin-tools>
     <version.maven-project-info-reports-plugin>3.4.3</version.maven-project-info-reports-plugin>
     <version.maven-release-plugin>3.0.1</version.maven-release-plugin>
     <version.maven-remote-resources-plugin>3.1.0</version.maven-remote-resources-plugin>
@@ -128,8 +125,7 @@ under the License.
     <version.maven-shade-plugin>3.4.1</version.maven-shade-plugin>
     <version.maven-site-plugin>3.12.1</version.maven-site-plugin>
     <version.maven-source-plugin>3.3.0</version.maven-source-plugin>
-    <version.maven-surefire-plugin>${surefire.version}</version.maven-surefire-plugin>
-    <version.maven-surefire-report-plugin>${surefire.version}</version.maven-surefire-report-plugin>
+    <version.maven-surefire>${surefire.version}</version.maven-surefire>
     <version.maven-war-plugin>3.3.2</version.maven-war-plugin>
   </properties>
 
@@ -138,7 +134,7 @@ under the License.
       <dependency>
         <groupId>org.apache.maven.plugin-tools</groupId>
         <artifactId>maven-plugin-annotations</artifactId>
-        <version>${version.maven-plugin-annotations}</version>
+        <version>${version.maven-plugin-tools}</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
@@ -211,7 +207,7 @@ under the License.
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-failsafe-plugin</artifactId>
-          <version>${version.maven-failsafe-plugin}</version>
+          <version>${version.maven-surefire}</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
@@ -262,12 +258,12 @@ under the License.
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-plugin-plugin</artifactId>
-          <version>${version.maven-plugin-plugin}</version>
+          <version>${version.maven-plugin-tools}</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-plugin-report-plugin</artifactId>
-          <version>${version.maven-plugin-report-plugin}</version>
+          <version>${version.maven-plugin-tools}</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


martin-g (approved):

pom.xml
```diff
@@ -89,22 +89,56 @@ under the License.
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <sourceReleaseAssemblyDescriptor>source-release</sourceReleaseAssemblyDescriptor>
     <gpg.useagent>true</gpg.useagent>
-    <minimalMavenBuildVersion>3.2.5</minimalMavenBuildVersion>
+    <minimalMavenBuildVersion>3.6.3</minimalMavenBuildVersion>
```
slawekjaranowski, line 92: We have separate issue for it: https://issues.apache.org/jira/browse/MPOM-399

pom.xml
```diff
@@ -89,22 +89,56 @@ under the License.
...
     <minimalJavaBuildVersion>1.8</minimalJavaBuildVersion>
     <maven.compiler.source>${maven.compiler.target}</maven.compiler.source>
     <maven.compiler.target>1.7</maven.compiler.target>
-    <surefire.version>3.1.0</surefire.version><!-- for surefire, failsafe and surefire-report -->
+    <surefire.version>3.1.2</surefire.version><!-- for surefire, failsafe and surefire-report -->
```
slawekjaranowski, line 96: Update should be done as separate PR and issue

pom.xml
```diff
@@ -89,22 +89,56 @@ under the License.
...
+    <version.maven-gpg-plugin>3.1.0</version.maven-gpg-plugin>
+    <version.maven-help-plugin>3.4.0</version.maven-help-plugin>
+    <version.maven-install-plugin>3.1.1</version.maven-install-plugin>
+    <version.maven-invoker-plugin>3.5.1</version.maven-invoker-plugin>
+    <version.maven-jar-plugin>3.3.0</version.maven-jar-plugin>
+    <version.maven-javadoc-plugin>3.5.0</version.maven-javadoc-plugin>
+    <version.maven-plugin-annotations>${maven.plugin.tools.version}</version.maven-plugin-annotations>
+    <version.maven-plugin-plugin>${maven.plugin.tools.version}</version.maven-plugin-plugin>
+    <version.maven-plugin-report-plugin>${maven.plugin.tools.version}</version.maven-plugin-report-plugin>
+    <version.maven-project-info-reports-plugin>3.4.5</version.maven-project-info-reports-plugin>
```
slawekjaranowski, line 122: Next version update

slawekjaranowski (commented):
Looks ok for me.

To think if we want separate PRs for other updates as I commented or link / create issues for it.

pom.xml
ctubbsii: I can back that out of here.

pom.xml
ctubbsii: Sure, no problem. I'll back it out.

gnodet (approved):

slachiewicz (approved):

pom.xml
```diff
@@ -97,14 +97,48 @@ under the License.
...
+    <version.maven-project-info-reports-plugin>3.4.3</version.maven-project-info-reports-plugin>
+    <version.maven-release-plugin>3.0.1</version.maven-release-plugin>
+    <version.maven-remote-resources-plugin>3.1.0</version.maven-remote-resources-plugin>
+    <version.maven-resources-plugin>3.3.1</version.maven-resources-plugin>
+    <version.maven-scm-plugin>2.0.1</version.maven-scm-plugin>
+    <version.maven-scm-publish-plugin>3.2.1</version.maven-scm-publish-plugin>
+    <version.maven-shade-plugin>3.4.1</version.maven-shade-plugin>
+    <version.maven-site-plugin>3.12.1</version.maven-site-plugin>
+    <version.maven-source-plugin>3.3.0</version.maven-source-plugin>
+    <version.maven-surefire-plugin>${surefire.version}</version.maven-surefire-plugin>
```
olamy, line 131: as the standard is now to use `<version.artifactId>` to contain version maybe this should be done in the other way to be consistent.
```
<version.maven-surefire-plugin>xxxx</version.maven-surefire-plugin>
<surefire.version>${version.maven-surefire-plugin}</surefire.version>
```

olamy (changes requested):

pom.xml
hboutemy: I see the logic, but this will break the configuration that people have done with `surefire.version` until now

then I think that the current reuse of `surefire.version` is a nicer migration path

hboutemy (approved):
LGTM

pom.xml
olamy: Not sure to understand what this will break as the property will be still there.
Any example how this will break something?

slawekjaranowski: `surefire.version` is also used in other place, like 
```
    <version.maven-failsafe-plugin>${surefire.version}</version.maven-failsafe-plugin>
```

olamy: Frankly if you think it's ok. I have no intention to block anything. 
I just find it not consistent.

slawekjaranowski (approved):

pom.xml
ctubbsii: @olamy,

The way it currently is, a user can consistently use the new properties, and things will work as expected.
Or, if they don't know about the new properties, but are using the old one, then it will still work as expected.

If we were to reverse these, we'd get consistency of presentation in the POM, but it would hurt the user's experience:

1. First, we'd have to figure out how to deal with the failsafe plugin's version, because `surefire.version` is a special case due to it controlling 3 plugins. (`maven.plugin.tools.version` is a similar special case). If we reverse the order, then what would we do about the other 2 affected by that same property? It's not clear.
2. Second, any existing `surefire.version` set in the user's POM would be irrelevant and ignored if the user didn't know to override it using the new property. The entire value of `surefire.version` as a property to control multiple plugins would be completely irrelevant (unless we also changed the `<plugin><version/>` value to use the other property, in which case we'd have inconsistency in the `<plugins>` section there instead of in the `<properties>` section, and we wouldn't actually solve the inconsistency you're pointing out.

The current implementation I have preserves a consistent user experience. Even if it looks a little inconsistent and out of place in the POM, there's a really good reason for it.

olamy: > @olamy,
> 
> The way it currently is, a user can consistently use the new properties, and things will work as expected. Or, if they don't know about the new properties, but are using the old one, then it will still work as expected.
> 
> If we were to reverse these, we'd get consistency of presentation in the POM, but it would hurt the user's experience:
> 
> 1. First, we'd have to figure out how to deal with the failsafe plugin's version, because `surefire.version` is a special case due to it controlling 3 plugins. (`maven.plugin.tools.version` is a similar special case). If we reverse the order, then what would we do about the other 2 affected by that same property? It's not clear.
> 2. Second, any existing `surefire.version` set in the user's POM would be irrelevant and ignored if the user didn't know to override it using the new property. The entire value of `surefire.version` as a property to control multiple plugins would be completely irrelevant (unless we also changed the `<plugin><version/>` value to use the other property, in which case we'd have inconsistency in the `<plugins>` section there instead of in the `<properties>` section, and we wouldn't actually solve the inconsistency you're pointing out.
> 
> The current implementation I have preserves a consistent user experience. Even if it looks a little inconsistent and out of place in the POM, there's a really good reason for it.

as long as a change is documented users can adjust their usage. If we follow this argument nothing can be changed for life...
we could have
`<version.maven-failsafe-plugin>${version.maven-surefire-plugin}</version.maven-failsafe-plugin>`

Anyway I already don't like this but would prefer `<maven-surefire-plugin.version>` 
But as a majority prefers something else just do it, I can live with that.
It's just a matter of preference but don't waste your time trying to convince me ;)

slawekjaranowski: I agree with @olamy we can have:

```
<version.maven-failsafe-plugin>${version.maven-surefire-plugin}</version.maven-failsafe-plugin>
...
<version.maven-plugin-report-plugin>${version.maven-plugin-plugin}</version.maven-plugin-report-plugin>
```

and so on

michael-o: `version.maven-plugin-tools` and `version.maven-surefire`, no?

slawekjaranowski: I see that we want to support backward compatibility for old properties `surefire.version` and `maven.plugin.tools.version`

yes - new property for aggregate version for many items should follow new conventions,

so we can have

```
<surefire.version>3.1.0</surefire.version> <!-- old property -->
<version.maven-surefire>${surefire.version}</version.maven-surefire> <!-- new aggregate  property -->
<version.maven-surefire-plugin>${version.maven-surefire}</version.maven-surefire-plugin>
<version.maven-failsafe-plugin>${version.maven-surefire}</version.maven-failsafe-plugin>
```

michael-o: I would not introduce new split proprties, because either Surefire entirely is updated or none. It does not make sense to mix and match. Same for Plugin Tools. It just needs to be documented.

slawekjaranowski: @ctubbsii, @michael-o - I updated docs and use one property for such groups

michael-o (approved):

slachiewicz (approved):

jira-importer:
Resolve #489

PR #159, opened 2023-06-11
[MPOM-390] Upgrade to Maven Fluido Skin 1.11.2

slawekjaranowski:


michael-o (approved):
Attention: This upgrade will prevent everyone from using Maven Site Plugin 4+ unless this skin is also bumped to 2 which is good: explicit failure over implicit issues.

jira-importer:
Resolve #354

PR #173, opened 2023-10-26
[MPOM-440] Add maven-checkstyle-plugin to pluginManagement

slawekjaranowski:


slachiewicz (approved):

pom.xml
```diff
@@ -181,6 +182,11 @@ under the License.
           <artifactId>maven-clean-plugin</artifactId>
           <version>${version.maven-clean-plugin}</version>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-checkstyle-plugin</artifactId>
+          <version>${version.maven-checkstyle-plugin}</version>
+        </plugin>
```
ctubbsii, line 189: The only problem with this is that the maven-checkstyle-plugin often uses a very old version of checkstyle, so to be useful to users, they still often have to define the plugin to override the dependency, as in:

```xml
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-checkstyle-plugin</artifactId>
        <dependencies>
          <dependency>
            <groupId>com.puppycrawl.tools</groupId>
            <artifactId>checkstyle</artifactId>
            <version>10.12.2</version>
          </dependency>
        </dependencies>
        <executions>
          <execution>
            <id>check-style</id>
            <goals>
              <goal>check</goal>
            </goals>
            <configuration>
              <!-- execution configuration here -->
            </configuration>
          </execution>
        </executions>
      </plugin>
```

slawekjaranowski: yes it is true.
But newer checkstyle required JDK 11+, so in this place we can not make such assumptions.
child project also define specific plugin configurations.
So here only plugin version is reasonable.

slachiewicz: Maybe we can define checkstyle version as additional property?

slawekjaranowski: I don't want to much complicate the ASF parent pom.
So I think that managing plugin version is good enough.

jira-importer:
Resolve #385

PR #188, opened 2024-01-15
[MPOM-264] Set "maven.compiler.release" on JDK 9+

kwin:


kwin:
FTR: This is a simplified version of #48 which was declined back then. Nowadays the profile is even used within Maven Parent: https://issues.apache.org/jira/browse/MPOM-447

pom.xml
```diff
@@ -93,8 +93,9 @@ under the License.
     <gpg.useagent>true</gpg.useagent>
     <minimalMavenBuildVersion>3.6.3</minimalMavenBuildVersion>
     <minimalJavaBuildVersion>1.8</minimalJavaBuildVersion>
+    <targetJavaVersion></targetJavaVersion>
     <maven.compiler.source>${maven.compiler.target}</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.target>7</maven.compiler.target><!-- use version numbers without the "1." prefix (supported since javac 8) -->
```
kwin, line 98: should we enforce with https://maven.apache.org/enforcer/enforcer-rules/requireProperty.html that this property value does no longer start with "1."?

slawekjaranowski: I'm not sure if `7` is supported .... maybe it is time to switch it to `8`

kwin: yes it is: back to 5 if javac 8 is used, https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javac.html#BHCDIFEE

kwin: javac 17 still supports 7, only javac21 only supports 8+.

pom.xml
```diff
@@ -403,6 +403,26 @@ under the License.
...
+                  <regexMessage>The value of property maven.compiler.target must not start with "1.", use for example "8" instead of "1.8"!</regexMessage>
+                </requireProperty>
+                <requireProperty>
+                  <property>maven.compiler.source</property>
+                  <regex>(?!1\.).*</regex>
+                  <regexMessage>The value of property maven.compiler.source must not start with "1.", use for example "8" instead of "1.8"!</regexMessage>
+                </requireProperty>
+              </rules>
+            </configuration>
+          </execution>
```
slawekjaranowski, line 425: I would like to not add next execution, simply if user has wrong value build will fail.

michael-o: I'll let it fail implicitly...

pom.xml
```diff
@@ -525,6 +545,17 @@ under the License.
       </build>
     </profile>
     <!-- END SNIPPET: release-profile -->
+    <profile>
+      <id>jdk9ornewer</id>
```
michael-o, line 549: Make is simple: `jdk9+`?

slawekjaranowski (commented):
We also should update documentation.
https://maven.apache.org/pom/asf/

pom.xml
kwin: I removed the enforcement of the property in https://github.com/apache/maven-apache-parent/pull/188/commits/8ab37cef72208e2b9667dbd298a446b16696ec66.

pom.xml
kwin: Done in https://github.com/apache/maven-apache-parent/pull/188/commits/8ab37cef72208e2b9667dbd298a446b16696ec66.

slawekjaranowski (approved):

michael-o (approved):

jira-importer:
Resolve #479

PR #189, opened 2024-01-15
[MPOM-265] Remove no longer used ".maven-apache-parent.marker" file

kwin:
This has been introduced in MPOM-255 for enabling a profile only when building the ASF parent itself.


slachiewicz (approved):

jira-importer:
Resolve #460

PR #192, opened 2024-01-23
Fix documented output directory

kwin:
Improve formatting, minor cleanup

### DIFF
--- a/src/site-docs/apt/index.apt.vm
+++ b/src/site-docs/apt/index.apt.vm
@@ -126,7 +126,7 @@ The <<<apache-release>>> Profile
 
    The profile includes the following plugins:
 
- * maven-assembly-plugin:
+ * <maven-assembly-plugin>:
     {{{/apache-resource-bundles/source-release/}org.apache.apache.resources:apache-source-release-assembly-descriptor:$context.get("version.apache-resource-bundles")}}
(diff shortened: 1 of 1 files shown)


src/site-docs/apt/index.apt.vm
```diff
@@ -136,27 +136,27 @@ The <<<apache-release>>> Profile
...
     a passphrase in <<<$\{gpg.passphrase\}>>>, presumably in your <<<settings.xml>>>.
 
- * checksum-maven-plugin: configured to create a checksum file(s) for source release as
+ * <checksum-maven-plugin>: configured to create a checksum file(s) for source release as
     required by {{{http://www.apache.org/dev/release-distribution#sigs-and-sums}Apache release distribution policy}}.
 
  []
 
-  When doing a release with maven-release-plugin, this creates files in <<<target/checkout/target>>> ready
+  When doing a release with <maven-release-plugin>, this creates files in <<<target>>> ready
```
slawekjaranowski, line 151: Released files are in `target/checkout/target` in `target` we have current project snapshot version.
So sentence was correct.

slawekjaranowski (changes requested):

src/site-docs/apt/index.apt.vm
kwin: You are right, this sentence refers only to releasing with m-r-p but not to the profile in general.

PR #210, opened 2024-04-12
[MPOM-483] Make a separate module for documentation

slawekjaranowski:
current: https://maven.apache.org/pom/asf/index.html
new: https://maven.apache.org/pom-archives/asf-LATEST/index.html


docs/pom.xml
```diff
@@ -82,11 +77,26 @@ under the License.
...
+        <artifactId>maven-install-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
ctubbsii, line 92: Can these skips be done with a property instead? Seems like it would be less config.

ctubbsii (commented):
Personally, I think creating a separate module for docs is confusing. From a project structure perspective, it looks like the docs are documentation for the docs module, and that the main module doesn't have docs, when it's the other way around. In general, I think I'd prefer keeping the site generation for the module in the same module that is being documented. Doing it this way seems to be not very "Maven-y".

docs/pom.xml
slawekjaranowski: ok, good idea

slawekjaranowski:
In ASF parent pom there is more trick about publishing documentation and publishing parent pom. 
For documentation perspective we need add some information to pom ... but we don't want to propagate it to child project.
So current we have a special site-pom.xml which must by manually edited during release process.
I trying to avoid such manual taks.
https://maven.apache.org/developers/release/parent-pom-release.html

docs/pom.xml
slawekjaranowski: done

jira-importer:
Resolve #403

PR #241, opened 2024-07-03
Disable matrix build on GitHub

slawekjaranowski:
Similar like on Jenkins - we use one build for simple project


michael-o:
What are the implications? I am not firm with this?

slawekjaranowski:
> What are the implications? I am not firm with this?

Save CO2 and time, we not need build matrix with many JDK and Maven versions

michael-o (approved):

slachiewicz (approved):

PR #251, opened 2024-08-16
By default target Java 8

### LINKED ISSUE #250: Set more reasonable default for property "minimalJavaBuildVersion"
Currently the property `minimalJavaBuildVersion` defaults to 1.8 (https://github.com/apache/maven-apache-parent/blob/b6149240303a5d8cd59413b1544125548e752d21/pom.xml#L99C6-L99C29).
This is not a reasonable default if property `maven.compiler.target` is set to something above 8.

I would propose to increase the default value of `maven.compiler.target` to `8` (https://github.com/apache/maven-apache-parent/blob/b6149240303a5d8cd59413b1544125548e752d21/pom.xml#L101C28-L101C29) and at the same time change the default value of `minimalJavaBuildVersion` to `${maven.compiler.target}`.

kwin:
Adjust minimalJavaBuildVersion to be in line with target bytecode version.

This closes #250

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,9 +96,9 @@ under the License.
     <sourceReleaseAssemblyDescriptor>source-release</sourceReleaseAssemblyDescriptor>
     <gpg.useagent>true</gpg.useagent>
     <minimalMavenBuildVersion>3.6.3</minimalMavenBuildVersion>
-    <minimalJavaBuildVersion>1.8</minimalJavaBuildVersion>
+    <minimalJavaBuildVersion>${maven.compiler.target}</minimalJavaBuildVersion>
     <maven.compiler.source>${maven.compiler.target}</maven.compiler.source>
-    <maven.compiler.target>7</maven.compiler.target><!-- use version numbers without the "1." prefix (supported since javac 5) -->
+    <maven.compiler.target>8</maven.compiler.target><!-- use version numbers without the "1." prefix (supported since javac 5) -->
     <surefire.version>3.3.0</surefire.version><!-- for surefire, failsafe and surefire-report -->
     <assembly.tarLongFileMode>posix</assembly.tarLongFileMode>
 


slawekjaranowski (approved):

ctubbsii:
I don't think it's very important to make the minimal Java version match the target version, except in JDK versions earlier than 9, where you also have to set the bootstrap class path to avoid incorrect cross-compilation. So, this change makes sense, so long as the minimalJavaBuildVersion is less than 11 (the first LTS version of JDK after 9 when the release flag was introduced that strictly enforces cross-compilation compatibility with the target version).

Since this POM is still dealing with versions earlier than 9, this change makes sense. However, I'm wondering if it might also make sense to bump the minimalJavaBuildVersion to 11, and move the `maven.compiler.release` property out of the JDK9 profile and into the main props in the POM. That would make it difficult to use this POM with JDK versions earlier than 11, but would greatly improve cross-compilation issues for projects that still want to target Java versions earlier than 9.

kwin:
> I don't think it's very important to make the minimal Java version match the target version, except in JDK versions earlier than 9,

I disagree as the target bytecode version is always (except when using toolchains) the lower bound for the build java version. Cross-compilation only works towards lower major version, but not towards higher major version.

The release flag is already correctly considered in the parent POM (in a [dedicated Java9 profile](https://github.com/apache/maven-apache-parent/blob/ded34a863110021191af2db5b772b665a2789394/pom.xml#L530-L540)), so in case something does not work as expected here please raise a new issue @ctubbsii.

ctubbsii:
> > I don't think it's very important to make the minimal Java version match the target version, except in JDK versions earlier than 9,
> 
> I disagree as the target bytecode version is always (except when using toolchains) the lower bound for the build java version. Cross-compilation only works for lower major versions, but not for targeting higher major versions.

I don't understand what you're disagreeing with. I already know that it's the lower bound and never questioned that. The only thing I was discussing was when it's important to be an exact match.

1. `targetVersion < buildVersion && buildVersion < 9 && bootstrapClasspathIsSet`: SAFE
2. `targetVersion < buildVersion && buildVersion < 9 && !bootstrapClasspathIsSet` (before this PR): UNSAFE
3. `targetVersion == buildVersion` (after this PR): SAFE
4. `targetVersion <= buildVersion && buildVersion >= 9 && releaseFlagIsUsed`: SAFE (`==` is "not very important"; `<` is sufficient)

> 
> The `maven-compiler-release` property is already correctly set in the parent POM (in a [dedicated Java9 profile](https://github.com/apache/maven-apache-parent/blob/ded34a863110021191af2db5b772b665a2789394/pom.xml#L530-L540)), so in case something does not work as expected here please raise a new issue @ctubbsii.

Yes, I know. I directly referenced that profile in what I wrote above. It works fine. I think I'm being misunderstood. All I'm suggesting (for discussion) is whether we should go even further than this PR does, and make the minimalJavaBuildVersion default to 11 (item 4 above). I can propose that in a new PR, but I thought I'd get feedback in a comment here first.

kwin:
For me

> I don't think it's very important to make the minimal Java version match the target version, except in JDK versions earlier than 9,

sounds different. For me the default should always match the target version (no matter which JDK you are building with)!

ctubbsii:
> For me
> 
> > I don't think it's very important to make the minimal Java version match the target version, except in JDK versions earlier than 9,
> 
> sounds different. For me the default should always match the target version (no matter which JDK you are building with)!

Some plugins require newer Java build versions than the project. I frequently find myself in a situation where my project targets 8, but the build quality Maven plugins require 11. Also, as I've pointed out in other places, JDK 11 actually enforces the JLS for Java 8 more strictly than JDK 8 does. So, even if you're targeting 8, you're better off building with 11 using the release flag set to 8, if you want a strictly compliant Java 8 application that is maximally portable (OpenJDK 8 and Oracle JDK 8 both let some internal APIs leak into the compile time classpath that are in violation of the JLS 8 spec).

kwin:
You are talking about edge cases, I am talking about a reasonable default. And the most reasonable thing for now is having the minimum version by default(!) being the target version. However this is not the suitable format for discussions. Feel free to open another issue for other topics.

PR #285, opened 2024-12-21
APT to Markdown

kwin:


docs/src/site/markdown/index.md.vm
```diff
@@ -0,0 +1,134 @@
...
+
+# Apache Software Foundation Parent POM
+
+This project consists of a POM that contains settings that are likely to be useful to any Apache project that is building and releasing code with Maven\. By using this project as a parent, a project gets these settings\.
+
+Most projects will want to override some of these settings\.
+
+The following is a list of the items configured in this POM:
+
+- Top\-level **project properties**:
```
slawekjaranowski, line 35: do we need all backslashes ...

kwin: no, but this file and all escape characters are mostly autogenerated from MarkdownSink, which generates them as it is much harder to only use escape characters when really needed. However they don’t do any harm except for making the markdown harder to read.

docs/src/site/markdown/index.md.vm
```diff
@@ -0,0 +1,134 @@
...
+    - **license**: [Apache\-2\.0](https://www.apache.org/licenses/LICENSE-2.0.txt)\.
+    - **organization**: $<a id="project.organization.name"></a>project\.organization\.name
```
slawekjaranowski, line 37: was:
```
  ** <<organization>>: ${project.organization.name}
```
should be:
```
**organization**:  $context.get("project.organization.name") 
```

or even
```
**organization**:  $project.organization.name 

```

docs/src/site/markdown/index.md.vm
```diff
@@ -0,0 +1,134 @@
...
+    On developer workstations it expects GnuPG agent to be available to get passphrases, while on fully unattended \(CI\-like\) workflows the use of `MAVEN_GPG_PASSPHRASE` environment variable use is recommended\.
+
+- checksum\-maven\-plugin  
+    Configured to create a checksum file\(s\) for source release as required by [Apache release distribution policy](http://www.apache.org/dev/release-distribution#sigs-and-sums)\.
+
+When doing a release with maven\-release\-plugin, this creates files in `target/checkout/target` ready to be copied to projects&apos; [Apache `/dist/` release distribution](http://www.apache.org/dev/release-distribution) directory:
+
+- `${artifactId}-${version}-source-release.[zip|tar.gz]`: the source release archive\(s\)
+- `${artifactId}-${version}-source-release.[zip|tar.gz].sha512`: their checksum\(s\)
+- `${artifactId}-${version}-source-release.[zip|tar.gz].asc`: their signature\(s\)
```
slawekjaranowski, line 97: `${artifactId}`, `${version}` should be escaped?

slawekjaranowski (changes requested):

slachiewicz:
Could we try converting the files to static Markdown format to have better editing experience using local/web tools?

docs/src/site/markdown/index.md.vm
kwin: Not necessary

kwin:
> Could we try converting the files to static Markdown format to have better editing experience using local/web tools?

Let’s separate that effort in a dedicated PR

docs/src/site/markdown/index.md.vm
```diff
@@ -0,0 +1,134 @@
+---
+title: Apache Software Foundation Parent POM
+author: 
+  - Benson Margulies
+  - Hervé Boutemy
+  - Karl Heinz Marbaise
+date: 2016-05-09
```
gnodet, line 7: Do we need the authors / date information?  Those are available in git history with more accuracy and much more details.

docs/src/site/markdown/index.md.vm
```diff
@@ -0,0 +1,134 @@
+---
```
gnodet, line 1: I wonder if we should use two commits to somehow force the git history to see this file as a rename of `index.apt.vm`.

docs/src/site/markdown/index.md.vm
kwin: This is orthogonal to the conversion.

docs/src/site/markdown/index.md.vm
kwin: This PR has two commits for exactly that reason

docs/src/site/markdown/index.md.vm
gnodet: Right, but if you don't plan to remove it, shouldn't you add your name to keep it up-to-date ?

kwin: No, as I haven't changed the content, just the format.

gnodet: Ok, as you wish.  I don't really care much.  Fwiw, I don't consider this PR change as trivial.

slawekjaranowski:
@slachiewicz @kwin 
done in
- https://github.com/apache/maven-apache-parent/pull/595

PR #292, opened 2025-02-02
add Reproducible Central Report

hboutemy:
see sample report https://maven.apache.org/plugins/maven-artifact-plugin/reproducible-central.html

### DIFF
--- a/docs/src/site/apt/index.apt.vm
+++ b/docs/src/site/apt/index.apt.vm
@@ -121,6 +121,8 @@ Apache Software Foundation Parent POM
    The value will be updated by Maven Release Plugin during releases. If a project wants to disable Reproducible Builds, just define
    the property value with any single non-numeric character.
 
+   In version 34, Reproducible Central Report is added to report on Reproducible Builds checks for the project and it dependencies.
+
  []
 
 The <<<apache-release>>> Profile
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,7 @@ under the License.
     <version.apache-resource-bundles>1.7</version.apache-resource-bundles>
     <version.checksum-maven-plugin>1.11</version.checksum-maven-plugin>
     <version.maven-antrun-plugin>3.1.0</version.maven-antrun-plugin>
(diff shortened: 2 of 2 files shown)


hboutemy:
uh, looking at example, it seems badges are now blocked by csp
I'm not an expert, help appreciated to explain what exactly causes that, what should be done (probably at maven.apache.org site? or can be done in the report html page independently from site?)

hboutemy:
ok, researching:
https://en.wikipedia.org/wiki/Content_Security_Policy
https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html

I suppose some more restrictive CSP have been configured ASF-wide, need to find a pointer...

hboutemy:
https://infra.apache.org/csp.html supposed to be become effective March 1, 2025.
Not sure this is what is currently causing the issue, but perhaps there is an intermediate step

hboutemy:
https://privacy.apache.org/policies/website-policy.html

> 4) Using Assets from other Domains
> Assets (JavaScript files or snippets, images, fonts, CSS, etc.) from other domains cannot be loaded. All assets need to be hosted on ASF servers.

this may be that one that has been enabled over the past month

niallkp:
> @niallkp do you confirm that maven.apache.org webserver csp has been updated during last month to enforce that "4. Using Assets from other Domains" restriction, please?

@hboutemy Infra were going to implement the CSP temporarily on 1st February (yesterday) for testing purposes - but its not supposed to go permanently live until 1st March 2025. Heres the email with the plan (I think you need to be logged in to PonyMail to see):

- https://lists.apache.org/thread/w34sd92v4rz3j28hyddmt5tbprbdq6lc

I don't know if _**Infra**_ have turned that on or not for testing - would need to ask them.

slawekjaranowski (changes requested):
First we need a confirm with INFRA tema ... so now I remove from next release proposition

slachiewicz:
Closing as no longer relevant

PR #304, opened 2025-03-12
Fix example format and position in index doc

slawekjaranowski:


docs/src/site/apt/index.apt.vm
```diff
@@ -218,4 +211,10 @@ Notices
 </build>
 +------+
 
+ * Since version <<34>> <<<repositories>>> section has been moved to the <<<use-apache-snapshots>>> profile.
+
+   Apache snapshot repository is now optional, and is only added when the profile is activated.
+
+   It can be activated by setting the <<<apache.snapshots>>> property or activating the <<<use-apache-snapshots>>> profile directly.
```
Bukama, line 218: setting the property to what? I guess `true`? but maybe it's "active", "enabled". We should mentioned it whats proper values.

Bukama (approved):
Approval as my comment is not directly related to the changes made in this PR (but could applied easily here too).

docs/src/site/apt/index.apt.vm
slawekjaranowski: simply use user property as `-Dsomething` set default value as `true`
I would like to left is as is to not write a long description ...

PR #533, opened 2025-07-28
Update minimum maven version to match current stable version (3.6.3 -> 3.9)

DomGarguilo:
According to https://maven.apache.org/docs/history.html:
> Maven 3.6.3 and before has now reached its end of life and is no longer supported in any way.

This PR updates the version to current stable version 3.9.11

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,7 +98,7 @@ under the License.
     <gpg.useagent>true</gpg.useagent>
     <!-- java version used to to set maven.compiler.* -->
     <javaVersion>8</javaVersion>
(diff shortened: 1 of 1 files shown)


ctubbsii (approved):
I like it... but because some people are behind on patch versions, I would probably just set it to 3.9.0 as minimum, unless there's a really really good reason why people should be forced to not use those earlier 3.9 versions.

slawekjaranowski (approved):

slachiewicz (approved):

slachiewicz (approved):

PR #569, opened 2026-02-13
Set source and target in jdk9+ profile

ctubbsii:
Use `javaVersion` to set `maven.compiler.source` and `maven.compiler.target`, in addition to `maven.compiler.release` in the `jdk9+` profile, because some plugins still use these properties, even when `maven.compiler.release` is set. So, it is still useful to have these set automatically by the single property, `javaVersion`.

### DIFF
--- a/docs/src/site/apt/index.apt.vm
+++ b/docs/src/site/apt/index.apt.vm
@@ -82,6 +82,8 @@ Apache Software Foundation Parent POM
 
 +------+
   <maven.compiler.release>\${javaVersion}</maven.compiler.release>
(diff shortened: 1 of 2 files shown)


ctubbsii:
> > because some plugins still use these properties, ...
> 
> which plugins needs it?

The main one that affected me was [net.revelc.code.formatter:formatter-maven-plugin](https://code.revelc.net/formatter-maven-plugin/plugin-info.html), but I'm certain there are others out there.

ctubbsii:
I'm going to close this PR, since I no longer think it's worth doing.

PR #579, opened 2026-05-17
Aligns the properties for Surefire

Bukama:


Bukama:
I was tempted to remove the property at all and label the PR as "breaking", but on the other site I'm not sure if its worth.

slawekjaranowski (approved):

ctubbsii (commented):
This change (and whatever previous change broke it) completely misunderstands the purpose of the properties.

The whole point of having the `surefire.version` property above the others was to use that one to sync up the others. The ones that start with `version.` are immediately followed by the specific artifact that they control. The ones above that, like the `surefire.version` one, were supposed to be the independent ones that control multiple, to keep them synchronized.

I will submit a new PR to fix this properly, since it looks like this is already merged and possibly already included in a release.

PR #584, opened 2026-06-21
add push-to-atr profile

hboutemy:
as planned and tested during first release in https://github.com/apache/tooling-atr-maven-plugin/blob/atr-maven-plugin-1.0.0-alpha-1/pom.xml#L492

### DIFF
--- a/docs/src/site/apt/index.apt.vm
+++ b/docs/src/site/apt/index.apt.vm
@@ -215,6 +215,15 @@ The <<<apache-release>>> Profile
  {{{https://github.com/apache/infrastructure-asfyaml/blob/main/README.md#project}<<<project>>> entry in <<<.asf.yaml>>>}}.
  Once project is ready, start the release in ATR before doing the staging release with Maven with <<<mvn release:perform -Ppush-to-atr>>>.
 
+ Remember that:
+
+ * ATR helps your VOTE tracking, but is optional and still in its alpha phase (soon {{{https://github.com/apache/tooling-trusted-releases/issues/512}beta}}),
+
+ * this <<<push-to-atr>>> profile is still alpha too, and optional too: you can upload files to ATR and compose your release by hand if anything goes unexpected:
+   feedback on success and failures is welcome.
+
+ []
+
 Settings Configuration
   
   You can have a look at the page in Apache website
--- a/docs/src/site/apt/index.apt.vm
+++ b/docs/src/site/apt/index.apt.vm
@@ -204,6 +204,17 @@ The <<<apache-release>>> Profile
 
  Those files also end up in the staging repository and Maven Central.
 
+* The <<<push-to-atr>>> Profile
+
+ Since release 39, <<<push-to-atr>>> profile is provided to help upload built release content to
+ {{{https://tooling.apache.org/}Apache Tooling}}'s Trusted Releases platform, and compose the release
+ using {{{https://apache.github.io/tooling-atr-maven-plugin/}<<<atr-maven-plugin>>>}}.
+
+ See ATR's {{{https://release-test.apache.org/tutorial}tutorial}} for a quick introduction: you'll have to create a
+ project, for example using
+ {{{https://github.com/apache/infrastructure-asfyaml/blob/main/README.md#project}<<<project>>> entry in <<<.asf.yaml>>>}},
+ and start the release in ATR before composing.
+
 Settings Configuration
   
   You can have a look at the page in Apache website
--- a/docs/src/site/apt/index.apt.vm
+++ b/docs/src/site/apt/index.apt.vm
@@ -210,10 +210,10 @@ The <<<apache-release>>> Profile
  {{{https://tooling.apache.org/}Apache Tooling}}'s Trusted Releases platform, and compose the release
  using {{{https://apache.github.io/tooling-atr-maven-plugin/}<<<atr-maven-plugin>>>}}.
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -526,6 +527,39 @@ under the License.
...
     </profile>
     <!-- END SNIPPET: release-profile -->
+    <profile>
+      <id>push-to-atr</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.tooling</groupId>
+            <artifactId>atr-maven-plugin</artifactId>
+            <version>${version.tooling.atr}</version>
```
slawekjaranowski: I would like to push version in pluginManagement

docs/src/site/apt/index.apt.vm
```diff
@@ -204,6 +204,17 @@ The <<<apache-release>>> Profile
...
+* The <<<push-to-atr>>> Profile
+
+ Since release 39, <<<push-to-atr>>> profile is provided to help upload built release content to
+ {{{https://tooling.apache.org/}Apache Tooling}}'s Trusted Releases platform, and compose the release
+ using {{{https://apache.github.io/tooling-atr-maven-plugin/}<<<atr-maven-plugin>>>}}.
+
+ See ATR's {{{https://release-test.apache.org/tutorial}tutorial}} for a quick introduction: you'll have to create a
+ project, for example using
+ {{{https://github.com/apache/infrastructure-asfyaml/blob/main/README.md#project}<<<project>>> entry in <<<.asf.yaml>>>}},
+ and start the release in ATR before composing.
```
slawekjaranowski: maybe - before upload

we need manually start a new release in ATR before upload files, or we can use `atr.createVersion`

docs/src/site/apt/index.apt.vm
```diff
@@ -204,6 +204,17 @@ The <<<apache-release>>> Profile
 
  Those files also end up in the staging repository and Maven Central.
 
+* The <<<push-to-atr>>> Profile
+
+ Since release 39, <<<push-to-atr>>> profile is provided to help upload built release content to
+ {{{https://tooling.apache.org/}Apache Tooling}}'s Trusted Releases platform, and compose the release
+ using {{{https://apache.github.io/tooling-atr-maven-plugin/}<<<atr-maven-plugin>>>}}.
```
slawekjaranowski, line 211: by default we only help to upload files other steps need to done in ATR before release start

pom.xml
hboutemy: uh? why different from everything else?
(I'm the one who fights usually for that: why now and only for that plugin?)

docs/src/site/apt/index.apt.vm
hboutemy: don't hesitate to add a commit: phrasing is better by reworking together, not ping-pong comments

docs/src/site/apt/index.apt.vm
hboutemy: same as previous:  don't get what action you expect from that comment

pom.xml
slawekjaranowski: If some of project wants to extends configuration will not need to declare version for plugin

hboutemy: oh, you mean keep the property "as usual", but split plugin version declaration vs execution

adding a commit would have saved us time and energy: we need to find easier ways to do PR review that is more collaborative coding than real just review and discuss against original code proposal

docs/src/site/apt/index.apt.vm
```diff
@@ -204,6 +204,17 @@ The <<<apache-release>>> Profile
...
+
+ See ATR's {{{https://release-test.apache.org/tutorial}tutorial}} for a quick introduction: you'll have to create a
```
dave2wave: The main project for a PMC is always present. If a DOAP file is present then the subproject should also exist. Suggest that the user check their committee at https://release-test.apache.org/committees

hboutemy: reworked instructions

hboutemy:
notice: staging resulting doc is visible at https://maven.apache.org/pom-archives/asf-LATEST/#the-push-to-atr-profile

pom.xml
```diff
@@ -526,6 +532,38 @@ under the License.
...
+                </goals>
+              </execution>
+              <execution>
+                <id>upload-to-atr</id>
+                <goals>
+                  <goal>upload</goal>
+                </goals>
+                <configuration>
+                  <files>
+                    <file>${project.build.directory}/${project.artifactId}-${project.version}-source-release.zip</file>
```
ctubbsii, line 556: I realize this is already merged and I'm late to review, but hard-coding the paths to upload in the plugin config seems like a problem if the hard-coded path is a zip file. zip files don't preserve POSIX execute bits on scripts and other contents, so many projects create a source-release tarball instead, which is configurable in the maven-release-plugin configuration in this parent POM. Any ATR plugin profile should also be similarly capable of being able to handle that.

pom.xml
```diff
@@ -526,6 +532,38 @@ under the License.
...
+                    <file>${project.build.directory}/${project.artifactId}-${project.version}-source-release.zip.sha512</file>
```
ctubbsii, line 557: Is this supposed to be the sha512 generated by the net.nicoulaj.maven.plugins:checksum-maven-plugin ? That plugin generates checksums for all artifacts with the classifier `source-release`, not a specific `source-release` zip file. This plugin config makes a lot of assumptions about the file name of the source-release artifact.

PR #587, opened 2026-06-29
#586: Use RAT0.18 and remove commons-lang3 configuration for JDK25

### LINKED ISSUE #586: Use RAT 0.18
### New feature, improvement proposal

Not sure which JDKs are supported, but RAT 0.18 raised the language level to 17 and recommends using JDK21 ....

ottlinger:
Enable RAT0.18.

As RAT0.18 needs at least JDK17 I'm not sure which repos contain the actual plugin configuration, so this is a first draft PR.

Triggered by https://github.com/apache/maven-apache-parent/pull/574 @slawekjaranowski @slachiewicz

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,7 +106,7 @@ under the License.
 
     <project.build.outputTimestamp>2026-06-25T18:51:10Z</project.build.outputTimestamp>
 
-    <version.apache-rat-plugin>0.16.1</version.apache-rat-plugin>
+    <version.apache-rat-plugin>0.18</version.apache-rat-plugin>
     <version.apache-resource-bundles>1.8</version.apache-resource-bundles>
     <version.checksum-maven-plugin>1.11</version.checksum-maven-plugin>
     <version.maven-antrun-plugin>3.2.0</version.maven-antrun-plugin>
@@ -345,16 +345,6 @@ under the License.
           <groupId>org.apache.rat</groupId>
           <artifactId>apache-rat-plugin</artifactId>
           <version>${version.apache-rat-plugin}</version>
-          <dependencies>
-            <!-- avoid - WARNING: Use of the three-letter time zone ID ... on JDK 25 -->
-            <!-- https://issues.apache.org/jira/browse/RAT-490 -->
-            <!-- TODO remove with next version of RAT -->
-            <dependency>
-              <groupId>org.apache.commons</groupId>
-              <artifactId>commons-lang3</artifactId>
-              <version>3.20.0</version>
-            </dependency>
-          </dependencies>
         </plugin>
         <plugin>
           <groupId>org.apache.tooling</groupId>


ctubbsii (approved):

ctubbsii (changes requested):
This can't be updated yet. See my comment on #574

slachiewicz (approved):

ctubbsii:
This is a duplicate of #574 and can be closed.

slawekjaranowski (approved):

pom.xml
ctubbsii, line 100:
```suggestion
    <javaVersion>17</javaVersion>
```

ctubbsii (changes requested):
This should not be merged unless the minimalJavaBuildVersion is also bumped to 17.

This PR should have be closed as a duplicate of #574, which had the more detailed explanation of this in the code review comments, but for some reason, that original issue was closed, and this newer one without the discussion was left open (and subsequently approved by folks who apparently hadn't read the problems with this change).

pom.xml
ctubbsii: Alternatively, the minimalJavaBuildVersion can be set to 17, and the javaVersion can stay on an older version. The rat plugin requires 17 to run.

slawekjaranowski: I would like to not change `javaVersion` or `minimalJavaBuildVersion`
here we only provide a version there is not execution.

We can document it, there are an options when project will execute plugin 
- set javaVersion or minimalJavaBuildVersion
- profile can also be used to execute on JDK 17+

ctubbsii: By not changing either of those properties, the default behavior when people use this parent POM is that executions of rat will not work.

Users will have to choose one of:

1. Override `minimalJavaBuildVersion` to at least 17 so that rat works, OR
2. Revert to an older Apache parent POM and manually update all the plugin versions that are managed by the parent POM with local overrides.

Option 1 is basically what I'm suggesting doing anyway. Option 2 is a terrible option.

My preference is that the `minimalJavaBuildVersion` default be updated to 17 in this PR, because that's the version that is required for all the plugins specified in this POM to actually work. So, it makes sense. Users don't have to update to this version of the parent POM, but those who choose to, but also want to build on an older JDK would only need to override this property, and the version properties for the rat plugin, which is much less effort than Option 2.

There is a third option:

3. The rat plugin maintainers can revert the changes they made to require Java 17 (these are relatively few), and release an update that works with Java 8.

I don't see a reason for them to do that, though. Java 8 is so old at this point, and there's no reason to build with it, since newer JDKs build Java 8 code just fine, provided the `maven.compiler.release` property is properly set, as this POM does for them.

slachiewicz:
I measured what the 0.16.1 → 0.18 move actually changes, using `maven-parent` as the subject because it is where a `rat-check` execution is actually bound. Two findings, and they point in opposite directions.

## The version bump is not behaviour-neutral, even with configuration untouched

Same POM, same `<excludes>`, only the plugin version changed:

| Plugin version | Files scanned | Unapproved |
|---|---|---|
| 0.16.1 | 13 | 0 |
| 0.18 | 28 | 0 |

0.18 scans 15 files that 0.16.1 skipped, because its default exclusion set changed. Everything extra was licensed here, so `maven-parent` stays green — but a project holding an unlicensed file that 0.16.1 happened to skip will fail on the bump for a reason unrelated to this PR. Worth a release-note sentence so maintainers check rather than assume.

This does not argue against merging. `apache-parent` declares rat only in `pluginManagement`, with no `<execution>` and no `<configuration>`, so nothing here executes rat and nothing changes for a child until it bumps its own parent version.

## Do not "fix" the deprecation warning by renaming the element

0.18 warns `Use of deprecated option 'exclude'`. The obvious response is to rename `<excludes>` to `<inputExcludes>`. That is not a rename. Measured on `maven-parent` under 0.18, with the same ten patterns and nothing else altered:

| Configuration | Files scanned | Result |
|---|---|---|
| `<excludes>` (deprecated) | 28 | 0 unapproved, BUILD SUCCESS |
| `<inputExcludes>` (renamed) | 4 | 0 unapproved, BUILD SUCCESS |

The 24 files that stop being checked are every file under `docs/` (17), `maven-extensions/`, `maven-plugins/`, `maven-shared-components/` and `maven-skins/` (2 each) — whole modules. The new option routes through the current engine, where `excludeSubProjects` (default `true`) takes effect; the deprecated one does not reach it.

The failure mode is the dangerous one: the license check stops covering most of the tree and still reports success. Nothing in the build output says coverage collapsed.

apache/maven#12402 shows the same trap from the other side — a hand-migration there dropped `.mvn/**` and rewrote `dir/**` as `dir`, producing 3298 unapproved files. That one at least went red.

## Suggested handling

- Merge this as-is. Old-style `<excludes>` keeps working under 0.18 with a warning, so no downstream project needs to change anything on the parent bump.
- In the release notes: rat 0.18 requires JDK 17 to execute (`major version: 61`; 0.17 is the last Java 8 line at `major version: 52` and can be pinned through `version.apache-rat-plugin`, carrying `commons-lang3` 3.20.0 with it), and 0.18 scans a slightly wider default set.
- Treat the `inputExclude` migration as separate, per project, and validate it by diffing RAT's scanned-file *list* before and after. The counter totals do not reveal this: both configurations above report `0 unapproved`.

For evidence across the core projects before this lands, apache/maven-parent#598 puts 0.18 behind a temporary `[17,)`-activated profile so the JDK 8 CI cell keeps 0.16.1.

One procedural note: the comment above suggesting this is a duplicate of #574 has the direction reversed. #574 is closed and unmerged — closed with "suppressed by #587". Closing this one would leave the estate on 0.16.1.

Commands to replay: `javap -v org/apache/rat/plugin/RatCheckMojo.class` from the 0.17 and 0.18 plugin jars; `mvn -N org.apache.rat:apache-rat-plugin:{0.16.1,0.18}:check` on `maven-parent` master, comparing `target/rat.txt`.

*This comment was created with AI assistance.*

slawekjaranowski:
ok, go with it ....

PR #595, opened 2026-08-06
Port the site documentation from APT to Markdown

slachiewicz:
Converts the site documentation under `src/site/apt` to Markdown. APT is a format only
Doxia understands; Markdown is already supported by `maven-site-plugin` out of the box,
so no POM change is needed.

Converted with `doxia-converter` and then cleaned up by hand: `<<<code>>>` becomes
backticks, `<<bold>>` becomes `**bold**`, `{{{url}text}}` becomes `[text](url)`, and the
ASF licence header becomes a single block HTML comment. The page interpolates Velocity
properties, so it stays a `.md.vm`.

The change is split into two commits: a pure rename, so that `git log --follow` and
`git blame` still reach the page's history, and then the conversion itself. Please merge
or rebase rather than squash, or the rename is collapsed and that history is lost again.

Verified by building the site before and after the change and comparing the generated
page: the visible text and the link targets are unchanged.

Part of the wider migration tracked in
https://github.com/apache/maven-doxia-converter/issues/139


docs/src/site/markdown/index.md.vm
```diff
@@ -0,0 +1,178 @@
...
+
+        You can override this property in your project POM to set different Java version for compilation.
+
+        The `javaVersion` property is used to set the `maven.compiler.*` properties, so they should not be set directly in your project POM.
+
+        They will be set automatically according to the JDK used for the project build:
+
+        - for JDK 8 and older
+
+            ```unknown
```
ctubbsii, line 46: These ```` ```unknown ```` lines should be ```` ```xml ````

slawekjaranowski: Fortunately, renders good enough. 😄

ctubbsii: Yeah, the only difference would be syntax highlighting, which improves readability. It's optional, but nice.

PR #597, opened 2026-08-27
[#360] Publish SBOM on ASF releases

### LINKED ISSUE #360: [MPOM-401] publish SBOM on ASF releases
**[Herve Boutemy](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=hboutemy)** opened **[MPOM-401](https://issues.apache.org/jira/browse/MPOM-401?redirect=false)** and commented

same as MPOM-346, but for whole ASF

see ASF draft strategy https://cwiki.apache.org/confluence/display/COMDEV/SBOM

---

**Affects:** ASF-29

**Issue Links:**
- [MPOM-346](https://issues.apache.org/jira/browse/MPOM-346) publish SBOM on release

mcc0nnell:
Fixes #360

Publish a CycloneDX SBOM for ASF releases using the existing push-to-atr profile.

This:

* uses CycloneDX Maven Plugin 2.9.3
* runs makeAggregateBom during package
* generates JSON only
* sets outputReactorProjects=false so multi-module builds produce one aggregate SBOM
* allows projects to disable generation with -Dcyclonedx.skip=true
* creates a SHA-512 checksum and signature for the SBOM
* uploads the SBOM, checksum, and signature to ATR alongside the source release artifacts
* does not add SBOM artifacts to the normal Maven Central deployment path

This keeps the current CycloneDX 1.6 behavior scoped to ATR while leaving broader publication for CycloneDX 1.7 / isExternal support.

### DIFF
--- a/docs/src/site/markdown/index.md.vm
+++ b/docs/src/site/markdown/index.md.vm
@@ -130,6 +130,8 @@ The `push-to-atr` Profile
 
 Since release 39, `push-to-atr` profile is provided to help upload built release content to [Apache Tooling](https://tooling.apache.org/)'s Trusted Releases platform, and compose the release using [`atr-maven-plugin`](https://apache.github.io/tooling-atr-maven-plugin/).
 
+When this profile is active, an aggregate CycloneDX SBOM is generated during the `package` phase. Only JSON is generated and `outputReactorProjects` is disabled, so multi-module builds produce one aggregate SBOM rather than one SBOM per module. The SBOM, its SHA-512 checksum, and its signature are uploaded to ATR alongside the source release artifacts. Projects can disable SBOM generation with `-Dcyclonedx.skip=true`.
+
 See ATR's [tutorial](https://release-test.apache.org/tutorial) for a quick introduction: check existing [committes in ATR](https://release-test.apache.org/committees), you may have to create a project or improve its configuration, for example using [`project` entry in `.asf.yaml`](https://github.com/apache/infrastructure-asfyaml/blob/main/README.md#project). Once project is ready, start the release in ATR before doing the staging release with Maven with `mvn release:perform -Ppush-to-atr`.
 
 Remember that:
--- a/pom.xml
+++ b/pom.xml
@@ -121,6 +121,7 @@ under the License.
     <version.apache-rat-plugin>0.18</version.apache-rat-plugin>
     <version.atr-maven-plugin>1.0.0-alpha-1</version.atr-maven-plugin>
     <version.checksum-maven-plugin>1.11</version.checksum-maven-plugin>
+    <version.cyclonedx-maven-plugin>2.9.3</version.cyclonedx-maven-plugin>
     <version.maven-antrun-plugin>3.2.0</version.maven-antrun-plugin>
     <version.maven-assembly-plugin>3.8.0</version.maven-assembly-plugin>
     <version.maven-checkstyle-plugin>3.6.0</version.maven-checkstyle-plugin>
@@ -368,6 +369,11 @@ under the License.
           <artifactId>atr-maven-plugin</artifactId>
           <version>${version.atr-maven-plugin}</version>
         </plugin>
+        <plugin>
+          <groupId>org.cyclonedx</groupId>
+          <artifactId>cyclonedx-maven-plugin</artifactId>
+          <version>${version.cyclonedx-maven-plugin}</version>
+        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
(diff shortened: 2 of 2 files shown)


hboutemy:
FYI, we did it in Maven for a few years to test and learn: https://github.com/apache/maven-parent/blame/master/pom.xml#L1406

result is that aggregate SBOM makes people happy "because we have a SBOM" (in Maven Central), but the result is not really usable for anything useful

I won't explain in details why it's not useful, but focus on the solution: with the release of CycloneDX 1.7 ( few months ago) and its `isExternal` field, we start having the info that will make such SBOMs with CycloneDX Maven Plugin really useful
= https://github.com/CycloneDX/cyclonedx-maven-plugin/pull/696

once this CycloneDX 1.7 will be out, I'll be full interested into publishing these SBOMs everywhere

with current CycloneDX 1.6 plugin 2.9.x, we're taking the wrong route to me, adding generated files to the sizze of Maven Central without real benefit

slawekjaranowski:
Today I hope that ART will consume it ...

- we can generate only json format
- we can skip generating in all project modules (set outputReactorProjects = false)
- make allow to disable in child project
- optionally skip attache to deployment

Now I see that is used in:
https://github.com/search?q=%22%3CartifactId%3Ecyclonedx-maven-plugin%3C%2FartifactId%3E%22+language%3A%22Maven+POM%22+org%3Aapache&type=code

slawekjaranowski:
or only add in push-to-atr profile with deployment to atr

mcc0nnell:
That makes sense.

I’ll move this to push-to-atr and keep it out of the normal Maven Central deployment path for now.

I can also trim it to JSON only, set outputReactorProjects=false, and make it easy for child projects to disable.

Then once CycloneDX 1.7 / isExternal support is in the plugin, we can revisit broader publication.

That does seem like the cleaner path.

slachiewicz:
I would prefer to try adding change to one or two close-to-be-released-via-atr projects before enforce it on all projects

slawekjaranowski:
> I would prefer to try adding change to one or two close-to-be-released-via-atr projects before enforce it on all projects

I'm working with ART on better SBOM support, and dit some of test release with SBOM
you can look at https://github.com/apache/tooling-trusted-releases/issues/1544

slawekjaranowski:
@ppkarwasz what do you think about publishing SBOM as default

slawekjaranowski:
We also have
- https://cwiki.apache.org/confluence/spaces/COMDEV/pages/217391474/SBOM
- https://cwiki.apache.org/confluence/spaces/SECURITY/pages/287607444/SBOM+Software+Bill+of+Materials

mcc0nnell:
Thanks — I think the cleanest scope here is ATR integration rather than treating the current CycloneDX 1.6 aggregate as the final SBOM publication model.

The branch is intentionally narrowed to push-to-atr: JSON only, one aggregate for the reactor, child-project opt-out, and explicit SBOM/checksum/signature upload for ATR. I agree that broader/default publication is a separate question and is better revisited with CycloneDX 1.7 / isExternal support and evidence from real ATR releases.

One detail I do not want to overclaim: makeAggregateBom attaches the SBOM by default, so with release:perform -Ppush-to-atr it may still participate in the normal deploy lifecycle. We should make that boundary explicit before describing this as ATR-only publication.

The intended scope is narrower: when a project opts into push-to-atr, preserve the generated SBOM as release evidence that ATR can ingest and evaluate. This PR does not need to claim that the current aggregate SBOM is a complete artifact-level dependency model or settle ASF-wide default publication.

hboutemy (changes requested):
no
2 main reasons: 
1. ATR is supposed to offer services to review SBOMs independently from release process: release is too late to start reviewing
2. I want to clarify what has to be published to ASF distribution area and be part of official release content (= why 1 has to be a prerequisite)

and in addition, as I wrote in ATR Slack channel, a few of us know that aggregated SBOM generated by CycloneDX Maven Plugin (that I maintain for a few year: I know what I'm talking about) are not usable in real life
but with the new CycloneDX 1.7 version and its implementation, we can change it: https://github.com/CycloneDX/cyclonedx-maven-plugin/pull/696

it's easy to publish generated bloat = the return of experience of a few years publishing to Maven Central.
The new CDX 1.7 update is the improvement that we need to at the end publish something useful.

If you want to be useful, please help improving the CycloneDX Maven Plugin CDX 1.7 support

ppkarwasz:
I am also opposed to this PR, because this parent POM produces **only** a source archive, and I have no idea what kind of SBOM should be produced for it.

The `cyclonedx-maven-plugin` produces `build` SBOMs for the binary artifacts of a build. It **does** have limitations, but it takes into account the **properties** and **active profiles** used in the build, and produces a result that is correct for that single build run.

It is not meant to describe a source archive, for which:

- multiple **profiles** can change the outcome of a build,
- dependency versions can be **ranges**. This is rare, but it is allowed.

For a `pre-build` / source SBOM I would expect to see different information:

- Vendored source code, with the copyright holder and license of that code.
- Bundled JavaScript: most source archives also contain the sources for the website.
- Dependencies that are conditional on properties and profiles.
- **Curated** license information: is a dependency `Apache-2.0 OR LGPL-2.1-only` or `Apache-2.0 AND LGPL-2.1-only`? In the latter case, can I produce a build that omits the LGPL-licensed part?

I honestly find licensing tedious, so I have not thought these aspects through in detail.

@jerpelea has integrated SBOM-generation in Apache NuttX, together with the ability for users to opt out at build time of the handful of source files that are **not** `Apache-2.0` licensed. The generated SBOM is a `build` SBOM, but since all source files are tagged with a `SPDX-Licence-Identifier`, this could be reused to generate a `pre-build` SBOM. This is something I would like to see for Java projects too, but it should be generated by the `apache-rat-plugin` rather than by the `cyclonedx-maven-plugin`.

To summarize, a source archive SBOM should contain:

- The files in the source tree that were **not** authored by the project: `mvnw`, JavaScript for the website, vendored code. I have an idea how this could be done.
- Dependencies for all profiles, which at this point are necessarily **external**, so that users can see the license implications.

This is a lot of work, and it is mostly useful for companies that rebuild our Java code (AWS, Google, Red Hat) and either use it internally or ship it to their clients. The majority of users will rely on the `build` SBOM of what we publish to Maven Central, and for now I think they should configure a `cyclonedx-maven-plugin` execution themselves, so they “own” each configuration option.

Issue #608, opened 2026-09-10
jdk9+ profile replaces a child's source and target with release 8


slachiewicz:
Since 36 (#543, #551) the `jdk9+` profile sets `maven.compiler.release` to `${javaVersion}`, which defaults to 8. Parents 32 to 35 derived it from `maven.compiler.target` instead, so overriding `target` in a child still took effect there. maven-compiler-plugin gives `release` precedence over `source`/`target` (plexus-compiler emits `--release` whenever the value is non-empty), so a child POM that declares `maven.compiler.source` and `maven.compiler.target` but no `release`, which is every child written against a parent before 32, is compiled for Java 8 on JDK 9+ regardless of its own settings.

Observed on flink-connector-parent moving from `org.apache:apache:20` to 39 with source 11 / target 17:

```
[INFO] Compiling 130 source files with javac [debug release 8] to target/classes
[ERROR] TransactionOwnership.java:[74,29] method orElseThrow in class java.util.Optional<T> cannot be applied to given types
```

The compile error is the good outcome. A child whose sources happen to be Java 8 compatible is silently downgraded to bytecode 52 instead.

Each available workaround has a cost:

- Setting `javaVersion` fixes `release`, but it is a single value. A child with different source and target levels, or with per-profile targets, cannot express that through `--release`.
- Blanking `maven.compiler.release` restores `-source`/`-target` for javac, but maven-javadoc-plugin fails on the empty string (apache/maven-javadoc-plugin#1376).
- Overriding the compiler plugin's `release` parameter in the child works, but nothing points a user at it.

A POM profile cannot see whether the child set `maven.compiler.source`, so the profile itself cannot be made conditional. What would help:

1. Call this out in the release notes and on the parent's site: a child that sets `maven.compiler.source`/`target` must also set `javaVersion` or override `maven.compiler.release`, or its levels are ignored.
2. Consider having the parent's enforcer execution fail when `maven.compiler.source` or `maven.compiler.target` differs from `javaVersion` on JDK 9+, so the mismatch fails loudly instead of downgrading silently.

Context: [FLINK-40630](https://issues.apache.org/jira/browse/FLINK-40630).

*This issue was created with AI assistance.*

PR #609, opened 2026-09-10
Set maven.compiler.release unconditionally and drop the JDK profiles

slachiewicz:
Related to #608. `maven.compiler.release` is now set from `javaVersion` unconditionally and the `jdk8` and `jdk9+` profiles are gone. Both consumers gate the option themselves: plexus-compiler only emits `--release` when the javac in use is 9 or newer, reading the version from the toolchain executable when forking, and the javadoc mojo checks its runtime version. That removes the toolchains flaw the old profile's comment admits (MNG-6943) and one moving part, with no behaviour change on JDK 9+.

The parent deliberately does not define `maven.compiler.source` or `target` next to `release`: maven-compiler-plugin 4.x forwards all three and javac rejects the combination, which is what #543 established and why #569 was withdrawn. This PR therefore does not fix the silent override in #608 on the 3.x plugin; the docs now say plainly that a project must override `javaVersion`, not `source`/`target`. A loud failure for the 3.x case would be an enforcer rule in a follow-up.

Verified with a child project against 40-SNAPSHOT on JDK 25, reading the javac command line:

| plugin | child sets | result |
|---|---|---|
| 3.16.0 on Maven 3.10.0-rc-1 | nothing | `--release 8` |
| 3.16.0 | `maven.compiler.target` 11 | `--release 8`, unchanged from 39 |
| 3.16.0 | `javaVersion` 17 | `--release 17` |
| 4.0.0-beta-5 on Maven 4.0.0-SNAPSHOT | nothing | success |
| 4.0.0-beta-5 | `maven.compiler.target` 11 | `option --target cannot be used together with --release`, unchanged from 39 |
| 4.0.0-beta-5 | `javaVersion` 17 | success |

Behaviour change only on JDK 8 builds: this POM no longer sets `source` and `target` there, so the compiler plugin's own default of 1.8 applies unless the project sets them. Measured on Zulu 1.8.0_504 with Maven 3.9.16, child setting nothing:

| parent | javac gets | class file |
|---|---|---|
| 39 | `-source 8 -target 8` | major 52 |
| this PR | `-source 1.8 -target 1.8` | major 52 |

The compiler plugin's page says 3.13.0 "converts" `release` to `source`/`target` on JDK 8; MCOMPILER-582 added only ITs and docs, and its IT configures all three, so the conversion is the defaults coinciding. A child with `javaVersion` 7 cannot reach the difference: the parent's `RequireJavaVersion` rule rejects JDK 8 for range `[7,)` on 39 as well. The docs module builds.

*This change was created with AI assistance.*


slachiewicz:
Branch rewritten. The previous version derived `release` from `target` and set all three properties, which fails on maven-compiler-plugin 4.x (see the table that was in the description). This version only sets `maven.compiler.release` from `javaVersion` and removes the two JDK profiles; the description is replaced accordingly.

*This comment was created with AI assistance.*

